PR #164, opened 2014-07-25
[FIX] Configuration

Th3Mouk:
InvalidConfigurationException: The child node "container_types" at path "sonata_block.profiler" must be configured.

Hello ! I guess the defaultValue() was added after isRequired() and so don't need to be maintained


Soullivaneuh:
@Th3Mouk Can you provide the context and the steps to reproduce this issue and what is the expected result?

So I can write some test after.

Soullivaneuh:
Closing because of no answer. Feel free to reopen it if you want to continue.

PR #185, opened 2015-01-30
Migrate namespace for InlineConstraint

rande:


rande:
@dbu @lsmith77 Hello, there is a WIP to remove a dependency to the AdminBundle. This is a major BC break in interface however as you can see it is pretty easy to fix.

What do you think ?

lsmith77:
the next version will be 2.0 for us anyway

rande:
@lsmith77 ok so it is time to make a BC break ;)

lsmith77:
yeah .. break away! :)

rande:
done https://github.com/sonata-project/SonataBlockBundle/commit/06fac2531ac872ea94a650a345331e14c4da9eb0

Issue #174, opened 2014-10-24
Method "currentAncestor" for object "Knp\Menu\MenuItem" does not exist


thundo:
Hello,

I'm getting this error at sonata user profile page.

    Method "currentAncestor" for object "Knp\Menu\MenuItem" does not exist in SonataBlockBundle:Block:block_side_menu_template.html.twig at line 29

Composer relevant packages:

    sonata-project/block-bundle              2.2.13             Symfony SonataBlockBundle
    knplabs/knp-menu                         v2.0.1             An object oriented menu library
    knplabs/knp-menu-bundle                  v2.0.0             This bundle provides an integration of the KnpMenu library

Thank you

lbrunet:
http://stackoverflow.com/questions/25793518/method-currentancestor-for-object-knp-menu-menuitem-does-not-exist-in-sonata

rande:
fixed

PR #188, opened 2015-03-09
fixed broken knp-menu call

pulse00:
The `side-menu` code seems to be incompatible with [`knp 2.x`](https://github.com/sonata-project/SonataBlockBundle/blob/master/composer.json#L27). This patch fixes it. Disclaimer: Taken from [stackoverflow](http://stackoverflow.com/a/25848817/965338).

Refs #174


rande:
duplicate #183

PR #190, opened 2015-04-07
Extra parameter removed

fgpdaniel:
I believe it's not necessary to include media parameter


rande:
It might be useful but not used. This is BC break so we keep it like this for now.

fgpdaniel:
Ok! Thanks rande.

PR #192, opened 2015-04-08
Any change of ErrorElement from AdminBundle to CoreBundle

ckmm:


rande:
no, the good solution will be to avoid installing >2.3 on BlockBundle for 2.2 release

ckmm:
yeah, but i need to install AdminBundle 2.3 CoreBundle with BlockBundle 2.2 and there is the problem with these use ... ErrorElement

PR #194, opened 2015-05-01
Fix deprecated addAllowedTypes arguments usage

Soullivaneuh:
Fixed deprecation:

> Calling the Symfony\Component\OptionsResolver\OptionsResolver::addAllowedTypes method with an array of options is deprecated since version 2.6 and will be removed in 3.0. Use the new signature with a single option instead.

Symfony 2.3+ BC kept.


Soullivaneuh:
Failed jobs are caused by `sudo pip install -r Resources/doc/requirements.txt` command. Can you please restart them?

PR #196, opened 2015-05-01
Fix deprecated Event::getName usage

Soullivaneuh:
Fixes deprecated:

> The Symfony\Component\EventDispatcher\Event::getName method is deprecated since version 2.4 and will be removed in 3.0. The event name can be received in the listener call instead.


rande:
If this working with sf 2.3  ?

Soullivaneuh:
Yes. We only change internal logic on the code. Should not be a problem.

BTW, test are green.

Do you want me to check coverage?

PR #202, opened 2015-06-10
[WIP] Add configureSettings to block service

### LINKED ISSUE #201: Deprecated OptionsResolverInterface on BlockServiceInterface
On `BlockServiceInterface`, we have this method:

```php
/**
 * Define the default options for the block
 *
 * @param OptionsResolverInterface $resolver
 */
public function setDefaultSettings(OptionsResolverInterface $resolver);
```

`OptionsResolverInterface` is deprecated since Symfony `2.7` and will be remove in `3.0`.

We have to fix this deprecation issue. I propose this way:

* For `sonata/block-bundle` 2.4 (next minor version)
 * Make a `AbstractBlockService` which will implements `BlockServiceInterface`
 * Add a method `configureSettings(OptionsResolver $resolver)` and deprecate `setDefaultSettings`.
 * Update sonata default services to match the new signature.
* For `sonata/block-bundle` 3.0 (next major version)
 * Remove `setDefaultSettings` and add `configureSettings` to `BlockServiceInterface`

@rande sounds good for you? Am I missing something?

Soullivaneuh:
`AbstractBlockService::configureSettings` fixes Symfony deprecation issue.
`BlockServiceInterface::setDefaultSettings` is now deprecated.

Fixes #201

### DIFF
--- a/Block/AbstractBlockService.php
+++ b/Block/AbstractBlockService.php
@@ -1,0 +1,40 @@
+<?php
+
+/*
+ * This file is part of the Sonata project.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\BlockBundle\Block;
+
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\OptionsResolver\OptionsResolverInterface;
+
+/**
+ * Class AbstractBlockService
+ *
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+abstract class AbstractBlockService implements BlockServiceInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function setDefaultSettings(OptionsResolverInterface $resolver)
+    {
+        $this->configureSettings($resolver);
+    }
+
+    /**
+     * Define the default options for the block
+     *
+     * @param OptionsResolver $resolver
+     */
+    public function configureSettings(OptionsResolver $resolver)
+    {
+    }
+}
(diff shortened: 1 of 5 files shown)


rande:
The method can be moved to the BaseService class and trigger an deprecated error.

Soullivaneuh:
> The method can be moved to the BaseService class and trigger an deprecated error.

All custom BlockService should always extends this class?

rande:
No it is not mandatory, but as it is stated in the documentation most of people is using it

Soullivaneuh:
If this is not mandatory we shoudl not assume it as a default class.

At least, I plan to have `BaseBlockService` extending `AbstractBlockService`.

This PR is not yet ready, let's review it later! ;-)

rande:
Actually there is no need to have ``AbstractBlockService`` class. Just update the interface and that it.

Soullivaneuh:
Well, and where should I put my `configureSettings`?

In `BlockServiceInterface` ? This will be BC break.

In `BaseBlockService` ? This is not a mandatory class to be extends so users could pass out the deprecated warning.

Or I didn't understand all the things.

rande:
You are right, I didn't catch the BC break :/

Soullivaneuh:
Yes. I'm doing the Symfony way.

As you can see, the new method is on the `AbstractType`: https://github.com/symfony/symfony/blob/a44ceb3156e549ffc17baaab7a206be39b636d4f/src/Symfony/Component/Form/AbstractType.php#L56
But not on the related interface: https://github.com/symfony/symfony/blob/a44ceb3156e549ffc17baaab7a206be39b636d4f/src/Symfony/Component/Form/FormTypeInterface.php#L80

The new method will be on interface only for 3.0: https://github.com/symfony/symfony/blob/282ea41cf07edf886766fa321f735c11177a67c1/src/Symfony/Component/Form/FormTypeInterface.php#L76

rande:
@Soullivaneuh yes I get that now, well done ;)

PR #204, opened 2015-06-11
Add configureSettings to block service

### LINKED ISSUE #201: Deprecated OptionsResolverInterface on BlockServiceInterface

Soullivaneuh:
`AbstractBlockService::configureSettings` fixes Symfony deprecation issue.
`BlockServiceInterface::setDefaultSettings` is now deprecated.

Fixes #201

@rande Please don't merge it now! :-)

### DIFF
--- a/Block/BlockContextManager.php
+++ b/Block/BlockContextManager.php
@@ -19,6 +19,9 @@ use Symfony\Component\OptionsResolver\Exception\ExceptionInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 
+/**
+ * Class BlockContextManager
+ */
 class BlockContextManager implements BlockContextManagerInterface
 {
     protected $blockLoader;
@@ -37,6 +40,14 @@ class BlockContextManager implements BlockContextManagerInterface
     protected $logger;
 
     /**
+     * Used for deprecation check on {@link resolve} method.
+     * To be removed in 3.0 with BC system.
+     *
+     * @var array
+     */
+    private $reflectionCache;
+
+    /**
      * @param BlockLoaderInterface         $blockLoader
      * @param BlockServiceManagerInterface $blockService
      * @param array                        $cacheBlocks
@@ -45,10 +56,11 @@ class BlockContextManager implements BlockContextManagerInterface
     public function __construct(BlockLoaderInterface $blockLoader, BlockServiceManagerInterface $blockService,
         array $cacheBlocks = array(), LoggerInterface $logger = null
     ) {
-        $this->blockLoader  = $blockLoader;
-        $this->blockService = $blockService;
-        $this->cacheBlocks  = $cacheBlocks;
-        $this->logger       = $logger;
+        $this->blockLoader     = $blockLoader;
+        $this->blockService    = $blockService;
+        $this->cacheBlocks     = $cacheBlocks;
+        $this->logger          = $logger;
+        $this->reflectionCache = array();
     }
 
     /**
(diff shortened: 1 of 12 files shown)


Soullivaneuh:
I have to check against built-in services.

Block/BlockContextManager.php
```diff
@@ -214,4 +212,31 @@ protected function setDefaultExtraCacheKeys(BlockContextInterface $blockContext,
...
+    private function resolve(BlockInterface $block, $settings)
+    {
+        $optionsResolver = new OptionsResolver();
+
+        $this->configureSettings($optionsResolver, $block);
+
+        $service = $this->blockService->get($block);
+        $service->setDefaultSettings($optionsResolver, $block);
+
+        $reflector = new \ReflectionMethod($service, 'setDefaultSettings');
```
rande: I think this should be compute once with a CompilerPass to avoid performance impact on using the reflection API.

Soullivaneuh: How a `CompilerPass` will help to detect and throw deprecation errors?

I did the same method as Symfony. Please see: https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Form/ResolvedFormType.php#L201-L207

rande: You can store an array with the result of the reflection and still send a trigger. Or at least store the result for one service to avoid doing the same over and over if a Page is composed of the same block.

Soullivaneuh: To be honest, I have some difficulties to see what you want.

Can you maybe propose a PR on this one to show me?

rande: Something like this.

```
$this->reflectionCache = [
 'class\path' => ['isOldOverwritten' => true|false, 'isNewOverwritten' => true|false
];


$class = get_class($service);
if ($this->reflectionCache[$class]['isOldOverwritten'] && !$this->reflectionCache[$class]['isNewOverwritten']) {
        trigger_error('The Sonata\BlockBundle\Block\BlockServiceInterface::setDefaultSettings() method is deprecated since version 2.3 and will be removed in 3.0. Use configureSettings() instead. This method will be added to the BlockServiceInterface with SonataBlockBundle 3.0.', E_USER_DEPRECATED);
 }
```

rande: The reflectionCache cache attribute can either be compute once at with a CompilerPass or at runtime.

Soullivaneuh: Ok I'll try. :+1:

Soullivaneuh:
@rande Reflection cache is on the place. :+1:

rande:
Did you plan to add unit test ?

Soullivaneuh:
@rande Unit test are already here for those classes.

I'm missing something?

Soullivaneuh:
Tests updated to fix deprecation notice errors.

AFAIK, the coverage still the same.

Block/BlockContextManager.php
```diff
@@ -37,6 +40,14 @@ class BlockContextManager implements BlockContextManagerInterface
     protected $logger;
 
     /**
+     * Used for deprecation check on {@link resolve} method.
+     * To be removed in 3.0 with BC system.
+     *
+     * @var array
+     */
+    private $reflectionCache;
```
sagikazarmark: There is no need to cache reflection objects. See this benchmark: https://gist.github.com/mindplay-dk/3359812

sagikazarmark:
@Soullivaneuh PR sent for all the remaining open issues

PR #217, opened 2015-07-31
Prepare for new toolbar

WouterJ:
Prepare for the new toolbar design: symfony/symfony#15160

sf-sonata-block-toolbar

I created the SVG from the PNG logo, if you have a proper SVG file, it should be used instead.


Soullivaneuh:
Is that BC with older toolbar too?

WouterJ:
@Soullivaneuh yes, the `profiler_markup_version` variable defines the version of the toolbar. The older toolbar has no such variable, so it's set to 1 (using the `default` filter), the upcoming version of the toolbar has version 2.

Using this variable, you can support different versions of the toolbar. E.g. when version 3 comes out, you have:

    {% if profiler_markup_version == 1 %}
        {# 2.7- #}
    {% elseif profiler_markup_version  == 2 %}
        {# 2.8 #}
    {% else %}
        {# latest #}
    {% endif %}

Soullivaneuh:
Okay nice idea. But just major versions?

Soullivaneuh:
Note: Will review this when symfony/symfony#15160 will be merged.

WouterJ:
@Soullivaneuh toolbar markup version is not bound to Symfony's version. If "3", it just means it's the 3th design of the toolbar (that can already happen in 3.1 for instance).

Soullivaneuh:
I'm not talking about Symfony version but semantic versioning itself. ;-)

WouterJ:
Ah, in that case: Yes :) (there aren't going to be much redesigns that require markup changes).

Soullivaneuh:
sonata_block_debug_bar

Two things:

* Sonata logo seems to be cropped.
* Maybe we could try to get the same style like other section. Logo + number square. What do you think?

Soullivaneuh:
And for the second point, maybe we could put a more explicit icon that Sonata logo.

Like this one for example: http://fontawesome.io/icon/cubes/

ping @rande.

rande:
@WouterJ yep you can change the icon as @Soullivaneuh suggest.

Soullivaneuh:
@WouterJ SVG format could be found here: https://github.com/FortAwesome/Font-Awesome/blob/master/fonts/fontawesome-webfont.svg

rande:
ok to test

rande:
ok to test

rande:
ok to test

WouterJ:
I like the suggestion of changing the icon to something more describing. I changed it to another icon than the one suggested, as I find the style of this icon more consistent with the other icons and it looks better when it's just 24x24. This is also the icon used on the sonata-project.org website.

Btw, @Soullivaneuh, the official sonata project logo *is* cropped: https://sonata-project.org/bundles/sonatageneral/images/logo.png

rande:
Ok for the logo

Soullivaneuh:
> Btw, @Soullivaneuh, the official sonata project logo is cropped: https://sonata-project.org/bundles/sonatageneral/images/logo.png

Yes, and it's not adapted for the toolbar IMHO.

OK for the logo too, but I would like you to remove "blocks" and setup number rendering as other components:

sonata_block_icons

Like doctrine does.

Resources/views/Profiler/block.html.twig
```diff
@@ -1,7 +1,10 @@
 {% extends 'WebProfilerBundle:Profiler:layout.html.twig' %}
 
 {% block toolbar %}
+    {% set profiler_markup_version = profiler_markup_version|default(1) %}
+
     <div class="sf-toolbar-block">
+        {% if profiler_markup_version == 1 %}
         <div class="sf-toolbar-icon">
```
Soullivaneuh: Should be indented.

Resources/views/Profiler/block.html.twig
```diff
@@ -9,6 +12,16 @@
                 {% if collector.events|length > 0 %}<strong>*</strong>{% endif %}
             </a>
         </div>
+        {% else %}
+        <a href="{{ path('_profiler', { 'token': token, 'panel': name }) }}">
```
Soullivaneuh: Should be indented.

Resources/views/Profiler/block.html.twig
```diff
@@ -18,12 +31,16 @@
                 <b>Containers</b>
                 <span>{{ collector.containers|length }}</span>
             </div>
+            {% if profiler_markup_version == 1 %}
+            {# don't show the total number of blocks in the info in the new design,
```
Soullivaneuh: Should be indented.

WouterJ:
Thanks @Soullivaneuh, I've fixed the remaining issues.

Please note that the number shown is already equal to the other numbers. The doctrine panel, as shown in each screenshot in this PR, uses the old style. You can see the new styles here: https://github.com/symfony/symfony/pull/15160 I've removed "blocks" though.

Soullivaneuh:
Thank you @WouterJ.

WouterJ:
Thanks @Soullivaneuh for your review & merge!

PR #226, opened 2015-10-01
renamed cookbooks folder to cookbook (singular)

OskarStark:


OskarStark:
@rande i know this will change the frontend url, but how to handle this case?

just change it or is there a way to define a map for obsolete urls?

OskarStark:
rebased

Issue #229, opened 2015-10-07
why does we wrote "Blocks*" instead of "Blocks" in the Profiler?


OskarStark:
![bildschirmfoto 2015-10-05 um 13 11 31](https://cloud.githubusercontent.com/assets/995707/10337092/c7726cda-6cfd-11e5-97dc-0d3097a49d7c.png)

<img width="310" alt="bildschirmfoto 2015-10-07 um 14 12 50" src="https://cloud.githubusercontent.com/assets/995707/10337094/cb2f4852-6cfd-11e5-9cf5-c949e12fbab1.png">

can i change that @rande

OskarStark:
Ping @Soullivaneuh

Soullivaneuh:
That's a good question. Did you check the related code?

OskarStark:
No, not yet, but maybe its because of the events fired

![bildschirmfoto 2015-10-14 um 11 41 49](https://cloud.githubusercontent.com/assets/995707/10480089/96cf0bd6-7268-11e5-9044-4015a8190a67.png)

OskarStark:
If some blocks are rendered, the bubble should be green, would you like to fix this @mvhirsch ? :smile:

PR #231, opened 2015-10-14
colorize toolbar label if one or more blocks were rendered

mvhirsch:
@OskarStark  As mentioned in #229.

### DIFF
--- a/Resources/views/Profiler/block.html.twig
+++ b/Resources/views/Profiler/block.html.twig
@@ -8,7 +8,7 @@
             <div class="sf-toolbar-icon">
                 <a href="{{ path('_profiler', { 'token': token, 'panel': name }) }}">
                     {# fake image span #}<span style="width:0px; height: 28px; vertical-align: middle;"></span>
-                    <span class="sf-toolbar-status">{{ collector.getTotalBlock() }}</span> blocks
+                    <span class="sf-toolbar-status {% if collector.getTotalBlock() > 0 %}sf-toolbar-status-green{% endif %}">{{ collector.getTotalBlock() }}</span> blocks
                     {% if collector.events|length > 0 %}<strong>*</strong>{% endif %}
                 </a>
             </div>


Soullivaneuh:
Could we have a before/after screenshot?

Please take a look to SF 2.8 toolbar too.

OskarStark:
bildschirmfoto 2015-10-14 um 16 05 04
bildschirmfoto 2015-10-14 um 16 05 08

before it was grey :smile: 

@Soullivaneuh i would like to merge this PR as it is.

IMO we have to check and open a separate issue/PR for adjustments for 2.8 toolbar

Thank you @mvhirsch

Soullivaneuh:
> IMO we have to check and open a separate issue/PR for adjustments for 2.8 toolbar

No. 2.8 compatibility is already here. We can't accept broken PR for that. Both need to be handled.

OskarStark:
Ok, but i don't understand how adding a class name to a span could broke the panel...

can you explain more?

Soullivaneuh:
See this line: https://github.com/sonata-project/SonataBlockBundle/pull/231/files#diff-8d1d2fb5d440bda24d37396fb7790e89R7

Behavior should be the same. So we have to check this.

OskarStark:
good catch @Soullivaneuh 

@mvhirsch can you please also handle the 2.8 version?

mvhirsch:
@OskarStark @Soullivaneuh I don't think we need to colorize in the new symfony toolbar.
We should only colorize warnings/errors according to this blog entry:

http://symfony.com/blog/new-in-symfony-2-8-redesigned-web-debug-toolbar

> [...] Another important difference is the way errors and warnings are displayed. Previously we displayed just a small green/yellow/red icon, whereas now we change the entire panel background to easily spot when some error happens [...]

mvhirsch:
What do you think @OskarStark @Soullivaneuh?

OskarStark:
:+1: @mvhirsch

Soullivaneuh:
I don't know if we really should colorize it. See doctrine request, no colors.

Green color seems to be used for successful stuff like ajax request and more.

Not sure if this should be applied here.

ping @rande @sonata-project/contributors

Soullivaneuh:
According to the new [Sonata version management](https://github.com/sonata-project/SonataAdminBundle/issues/3053) and next major [release plan](https://github.com/sonata-project/SonataAdminBundle/issues/3731), this project has been refactored regarding branching and versioning.

If you see this message, your PR concerns a patch or a minor release and is not targeting the right branch.

So I'm closing this one, but don't see it as a refusal. If you think your work is still relevant and want to continue, feel free to reopen it on the right branch (e.g. the default one).

Regards.

PR #241, opened 2015-11-18
Use FakeTemplating in BlockService tests

core23:
This would be more helpfull when testing blocks.

You could reuse ``templating`` when testing the execute method https://github.com/sonata-project/SonataDashboardBundle/blob/master/Tests/Block/ContainerBlockServiceTest.php#L48-L51


OskarStark:
Thank you @core23

PR #245, opened 2015-12-02
improve CI integration

rande:


.travis.yml
```diff
@@ -17,36 +12,45 @@ cache:
...
+    - php: 5.3
+      env: SYMFONY_VERSION=2.8.*
+    - php: 5.4
+      env: SYMFONY_VERSION=2.8.*
     - php: 5.6
-      env: CS_FIXER=run
+      env: SYMFONY_VERSION=2.8.* CS_FIXER=run
     - php: 5.3
-      env: COMPOSER_FLAGS="--prefer-lowest"
+      env: SYMFONY_VERSION=2.3.* COMPOSER_FLAGS="--prefer-lowest"
```
Soullivaneuh: :-1: for this.

The prefer-lowest is overridden by `SYMFONY_VERSION` var.

Soullivaneuh:
I think at least one job should be run with no env var at all.

PR #249, opened 2015-12-05
Implemented a "sonata_block_exists" twig function

mweimerskirch:
This adds a "sonata_block_exists" twig function that checks if a given block type exists.

I also added "exists" methods to ``BlockHelper``, ``BlockHelperInterface``, ``BlockContextManager`` and ``BlockContextManagerInterface`` as well as ``BlockLoaderChain``, ``ServiceLoader`` and ``BlockLoaderInterface``. The latter 3 however already have a method called "support" which from the name sounds like it should do something similar, but it seems to always just return true. I didn't find out what that "support" method is really used for, so any help in understanding what it does is appreciated.

It is needed for https://github.com/sonata-project/SonataAdminBundle/pull/3469, but should be useful for other use cases as well.

### DIFF
--- a/Block/Loader/ServiceLoader.php
+++ b/Block/Loader/ServiceLoader.php
@@ -30,6 +30,18 @@ class ServiceLoader implements BlockLoaderInterface
     }
 
     /**
+     * Check if a given block type exists.
+     *
+     * @param string $type Block type to check for
+     *
+     * @return bool
+     */
+    public function exists($type)
+    {
+        return in_array($type, $this->types, true);
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function load($configuration)
--- a/Twig/Extension/BlockExtension.php
+++ b/Twig/Extension/BlockExtension.php
@@ -36,6 +36,9 @@ class BlockExtension extends \Twig_Extension
     public function getFunctions()
     {
         return array(
+            new \Twig_SimpleFunction('sonata_block_exists',
+                array($this->blockHelper, 'exists')
+            ),
             new \Twig_SimpleFunction('sonata_block_render',
                 array($this->blockHelper, 'render'),
                 array('is_safe' => array('html'))
--- a/Block/BlockContextManager.php
+++ b/Block/BlockContextManager.php
@@ -104,6 +104,18 @@ class BlockContextManager implements BlockContextManagerInterface
     }
 
     /**
+     * Check if a given block type exists.
+     *
+     * @param string $type Block type to check for
+     *
+     * @return bool
+     */
+    public function exists($type)
+    {
+        return $this->blockLoader->exists($type);
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function get($meta, array $settings = array())
--- a/Block/BlockLoaderChain.php
+++ b/Block/BlockLoaderChain.php
@@ -29,6 +29,24 @@ class BlockLoaderChain implements BlockLoaderInterface
     }
 
     /**
+     * Check if a given block type exists.
+     *
+     * @param string $type Block type to check for
+     *
+     * @return bool
+     */
+    public function exists($type)
+    {
+        foreach ($this->loaders as $loader) {
+            if ($loader->exists($type)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function load($block)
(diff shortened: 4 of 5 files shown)


Block/BlockContextManagerInterface.php
```diff
@@ -39,6 +39,13 @@ public function addSettingsByType($type, array $settings, $replace = false);
     public function addSettingsByClass($class, array $settings, $replace = false);
 
     /**
+     * @param string $type    Block type to check for
+     *
+     * @return bool
+     */
+    public function exists($type);
```
OskarStark: adding a method here is BC break

we can merge this without this method and add a separate PR for the next major IMO

ping @Soullivaneuh

Block/BlockLoaderInterface.php
```diff
@@ -16,6 +16,13 @@
 interface BlockLoaderInterface
 {
     /**
+     * @param string $type
+     *
+     * @return bool
+     */
+    public function exists($type);
```
OskarStark: same here

OskarStark:
I like this PR

ping @sonata-project/contributors

Block/BlockContextManagerInterface.php
mweimerskirch: IMHO it's not a good idea to remove the method from the interface. Yes, it is a BC break, but given that the method is called from within the bundle itself (from the Twig Extension), it should be marked as required in all implementations. But that's just my opinion.

dbu: adding a method to an interface is a BC break, and according to semver, this is not OK within minor versions. if i wrote my own implementation of this interface and update a minor version, i would suddenly see errors.

the BC way of doing this would be an additional interface that extends this interface to add the exists method, and the code that wants to do the exists checks would check if that additional interface is present. in the next major version, you could merge the two interfaces and simplify the code.

mweimerskirch: @dbu Ok, I can do that. What should I call it though? "BlockContextManagerExtendedInterface", "BlockContextManagerNewInterface", ...?

rande: @mweimerskirch just remove the function from the interface from now. It can be added in the next major version.

mweimerskirch:
I removed the methods from the two interfaces and squashed the commits

rande:
@mweimerskirch There is one CS to fix and we are ready to merge ;) Can you add a bit of documentation too ?

Thanks

mweimerskirch:
Should be ok now.

rande:
@mweimerskirch I think you php-cs-fixer is not up to date: https://travis-ci.org/sonata-project/SonataBlockBundle/jobs/95455163

There are more CS errors now :/

Block/Loader/ServiceLoader.php
```diff
@@ -30,6 +30,18 @@ public function __construct(array $types)
...
     /**
+     * Check if a given block type exists.
+     *
+     * @param string $type Block type to check for
+     *
+     * @return bool
+     */
+    public function exists($type)
+    {
+        return in_array($type, $this->types);
```
Soullivaneuh: strict comparison.

mweimerskirch:
Sorry, I changed the docblock at the last minute and forgot the trailing "."
Feeling a bit stupid now ;-)
I also added the strict check @Soullivaneuh suggested.

Twig/Extension/BlockExtension.php
```diff
@@ -36,6 +36,9 @@ public function __construct(BlockHelper $blockHelper)
     public function getFunctions()
     {
         return array(
+            new \Twig_SimpleFunction('sonata_block_exists',
```
Soullivaneuh: Please follow PSR-2 standard: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#46-method-and-function-calls

Soullivaneuh:
Too late. But this would be great to be corrected. ;-)

dbu:
erm. but won't we now get fatal errors when somebody is using a custom block type and the twig function is used? i would check method_exists and throw an exception.

also, there should be an issue for the milestone of the next major version to adjust the interface and remove the method_exists checks.

WouterJ:
@dbu's comment still applies, applications will no have hard errors when the block loader or block context manager doesn't have an `exists()` method (e.g. when using the Symfony CMF).

Also, instead of allowing the `$block` argument passed to `sonata_block_render`, this only allows passing the `$type`. This means it'll break with loaders using other block info, like the CMF loader.

At last, this is missing tests and docs. Making it hard to find out how to use it and how the loader should behave. What should the behaviour of this method be when the passed block info is not supported?

PR #257, opened 2016-04-19
fixes deprecated setDefaultSettings in BlockContextManager

mvhirsch:
Just a small deprecation fix internally.

I'm not sure about the `$service->setDefaultSettings($optionsResolver, $block);`.
Because technically we just can't update the Interface it depends on (BlockServiceInterface) since it can't be update until 3.0.

The other idea I just had: check if method exists on $service, if so call the "configureSettings"-method, otherwhise fallback to "setDefaultSettings".

What do you guys think?


greg0ire:
> The other idea I just had

I think this is the way to go if you want to support old versions.

mvhirsch:
@greg0ire okay, I just added it.

mvhirsch:
Is there any documentation that also needs an update?

greg0ire:
> Is there any documentation that also needs an update?

As you said, it's internal, so I'd say "no".

greg0ire:
Looks like you have some tests to fix.

mvhirsch:
@greg0ire fixed

greg0ire:
Please fixup the two last commits into the first one.

mvhirsch:
@greg0ire squashed them :-)

OskarStark:
LGTM @greg0ire can you please merge this?

OskarStark:
thank you @mvhirsch 🍺

PR #284, opened 2016-05-25
Exceptions in blocks are not critical

uwej711:
### Changelog

```markdown
### Changed
- Changed the log level on exceptions in BlockRenderer from critical to error, as critical log entries may trigger further actions, i.e. sending mails from monlog and similar. This makes even more sense as the default settings in the block bundle simply catch the exception.

```

### Subject

See the Changelog entry.

### TODO

- [ ] Fix tests

### DIFF
--- a/Block/BlockRenderer.php
+++ b/Block/BlockRenderer.php
@@ -93,7 +93,11 @@ class BlockRenderer implements BlockRendererInterface
             $response = $this->addMetaInformation($response, $blockContext, $service);
         } catch (\Exception $exception) {
             if ($this->logger) {
-                $this->logger->error(sprintf('[cms::renderBlock] block.id=%d - error while rendering block - %s', $block->getId(), $exception->getMessage()));
+                $this->logger->error(sprintf(
+                    '[cms::renderBlock] block.id=%d - error while rendering block - %s',
+                    $block->getId(),
+                    $exception->getMessage()
+                ), compact('exception'));
             }
 
             // reseting the state object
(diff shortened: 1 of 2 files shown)


greg0ire:
Please change the todo list to reflect what you think you actually need to do.

Block/BlockRenderer.php
```diff
@@ -93,7 +93,7 @@ public function render(BlockContextInterface $blockContext, Response $response =
             $response = $this->addMetaInformation($response, $blockContext, $service);
         } catch (\Exception $exception) {
             if ($this->logger) {
-                $this->logger->critical(sprintf('[cms::renderBlock] block.id=%d - error while rendering block - %s', $block->getId(), $exception->getMessage()));
+                $this->logger->error(sprintf('[cms::renderBlock] block.id=%d - error while rendering block - %s', $block->getId(), $exception->getMessage()));
```
greg0ire: Nice change! Could you also linebreak this, and add the exception object to the context argument of error ? Like this : 

```php
$this->logger->error(sprintf(
    '[cms::renderBlock] block.id=%d - error while rendering block - %s',
    $block->getId(),
    $exception->getMessage()
), compact('exception'));
```

Soullivaneuh:
Some modification should be made on changelog and commit message:

Changelog:

> The log level on exceptions in `BlockRenderer` from critical to error.

Commit title:

> Changed the log level on exceptions in BlockRenderer from critical to error

Commit body:

> As critical, log entries may trigger further actions, i.e. sending mails from monolog and similar. This makes even more sense as the default settings in the block bundle simply catch the exception.

Thanks.

Soullivaneuh:
BTW, this PR should be submitted agains 3.x, should not be?

greg0ire:
Indeed, I don't think it is a BC-break

uwej711:
See #285

PR #285, opened 2016-05-25
Changed the log level on exceptions in BlockRenderer from critical to…

uwej711:
### Changelog

```markdown
### Changed
- The log level on exceptions in `BlockRenderer` is decreased from critical to error.
```

### Subject

As critical, log entries may trigger further actions, i.e. sending mails from monolog and similar. This makes even more sense as the default settings in the block bundle simply catch the exception.


Soullivaneuh:
Thank you @uwej711!

PR #299, opened 2016-06-05
Removed unused container property from AbstractBlockServiceTest

core23:
### Changelog

```markdown
### Removed
- Removed `AbstractBlockServiceTest::$container`
```

### Subject

The property is and should never be used.


Soullivaneuh:
I'm not sure classes under `Tests` directory should follow the BC way...

cc @greg0ire @sonata-project/contributors

core23:
That's why I want to remove this in the next major.

greg0ire:
> I'm not sure classes under Tests directory should follow the BC way...

They are autoloadable, so they probably should :( … but anyway, this is not a very important change, so it can wait for next major release, there is no hurry

PR #302, opened 2016-06-10
Always use service id as block id

core23:
Depends on #344 

### Changelog

``` markdown
### Changed
- The block name is automatically set via `TweakCompilerPass`
```
### Subject

The id/name of a block is automatically set to the internal service id. This is the same mechanism like creating a admin. You can't declare the block name by yourself anymore.

### DIFF
--- a/Resources/translations/SonataBlockBundle.en.xliff
+++ b/Resources/translations/SonataBlockBundle.en.xliff
@@ -2,24 +2,24 @@
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" version="1.2">
     <file source-language="en" datatype="plaintext" original="file.ext">
         <body>
-            <trans-unit id="sonata.block.container">
-                <source>sonata.block.container</source>
+            <trans-unit id="sonata.block.service.container">
+                <source>sonata.block.service.container</source>
                 <target>Container</target>
             </trans-unit>
-            <trans-unit id="sonata.block.text">
-                <source>sonata.block.text</source>
+            <trans-unit id="sonata.block.service.text">
+                <source>sonata.block.service.text</source>
                 <target>Simple text</target>
             </trans-unit>
-            <trans-unit id="sonata.block.rss">
-                <source>sonata.block.rss</source>
+            <trans-unit id="sonata.block.service.rss">
+                <source>sonata.block.service.rss</source>
                 <target>RSS feed</target>
             </trans-unit>
-            <trans-unit id="sonata.block.menu">
-                <source>sonata.block.menu</source>
+            <trans-unit id="sonata.block.service.menu">
+                <source>sonata.block.service.menu</source>
                 <target>Menu</target>
             </trans-unit>
-            <trans-unit id="sonata.block.template">
-                <source>sonata.block.template</source>
+            <trans-unit id="sonata.block.service.template">
+                <source>sonata.block.service.template</source>
                 <target>Template</target>
             </trans-unit>
         </body>
--- a/DependencyInjection/Compiler/TweakCompilerPass.php
+++ b/DependencyInjection/Compiler/TweakCompilerPass.php
@@ -34,18 +34,8 @@ class TweakCompilerPass implements CompilerPassInterface
         foreach ($container->findTaggedServiceIds('sonata.block') as $id => $tags) {
             $definition = $container->getDefinition($id);
 
-            $arguments = $definition->getArguments();
-
             // Replace empty block id with service id
-            if (strlen($arguments[0]) == 0) {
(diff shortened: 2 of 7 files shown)


greg0ire:
helds => holds ?

Soullivaneuh:
> You don't have to declare the block name by yourself anymore.

But can you?

BC way possible?

Soullivaneuh:
BTW, I think what you did **is** BC, am I wrong?

greg0ire:
> BTW, I think what you did is BC, am I wrong?

The translations are changing :P

Soullivaneuh:
> The translations ids are changing :P

Maybe keep the old one? Or don't add `.service`?

core23:
> BTW, I think what you did is BC, am I wrong?

This is a BC break, because we have to change the translation keys OR the service ids. Furthermore, anybody who uses custom blocks can have new block names.

core23:
👍 for changing the service ids
👎 for changing the translation keys

Soullivaneuh:
> This is a BC break, because we have to change the translation keys OR the service ids.

Please see my previous sentences.

> Furthermore, anybody who uses custom blocks can have new block names.

Not sure to understand. Can you please elaborate?

> :+1: for changing the service ids
> :-1: for changing the translation keys

Well, some argument of why would be great. :wink:

core23:
>> The translations ids are changing 

> Maybe keep the old one? Or don't add .service?

IMO there is no reason for keeping the old translation ids, because they are never used after this change.

>> Furthermore, anybody who uses custom blocks can have new block names.

> Not sure to understand. Can you please elaborate?

For example [this bundle ](https://github.com/sonata-project/SonataMediaBundle/blob/3.x/Resources/config/block.xml#L10-L12) has custom blocks. Luckily they have matching block and service id, but we can't guarantee it for every bundle in the wordl.

Resources/translations/SonataBlockBundle.en.xliff
```diff
@@ -2,24 +2,24 @@
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" version="1.2">
     <file source-language="en" datatype="plaintext" original="file.ext">
         <body>
-            <trans-unit id="sonata.block.container">
-                <source>sonata.block.container</source>
+            <trans-unit id="sonata.block.service.container">
```
Soullivaneuh: IMO, just removing the `.service` will solve all our BC break issue. Am I wrong?

core23: If we just remove the `.service` in `<service id="sonata.block.service...` we have a BC break.

Soullivaneuh: image

core23: We can't handle this without a BC break. So which name do you prefer?

Soullivaneuh: `.service` looks better.

Soullivaneuh:
Could you please add a note about translation changes?

Soullivaneuh:
Note: IMO, this PR should be merged at the very end to manage possible new block service before 4.0 release.

core23:
Added a note the upgrade guide.

We should merge this ASAP in the master, so possible new blocks in this bundle could use the new mechanism.

Soullivaneuh:
> so possible new blocks in this bundle could use the new mechanism.

Absolutely not because new blocks in this bundle have to be added on 3.x, not master.

greg0ire:
> Absolutely not because new blocks in this bundle have to be added on 3.x, not master.

Should the base branch be changed then ? https://github.com/blog/2224-change-the-base-branch-of-a-pull-request

core23:
> Should [the base branch](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request) be changed then ? 

Great feature from GitHub 🎉 We can't do this in the stable branch, because this is a BC.

OskarStark:
Nice feature 🎉

OskarStark:
anything todo here @core23 ?

core23:
As @Soullivaneuh mentioned, this should be done just before releasing the 4.0 version.

Or we must care that the block name and service id is the same when creating new block inside the sonata bundles.

greg0ire:
Should we close this one in favor of #344 ?

core23:
No, this is the master change that will remove the BC and changes the block id's.

greg0ire:
Oh ok, great! But I think we shouldn't merge it until #344 is merged, and 3.x is merged into master, ok?

core23:
Rebased

greg0ire:
Thanks @core23 !

PR #310, opened 2016-06-29
Move FakeTemplating to public test namespace

core23:
I am targetting this branch, because this is a BC with a deprecation notice.

## Changelog

```markdown
### Deprecated
- Deprecate `Tests\Block\Service\FakeTemplating` in favor of `Test\Mock\MockTemplating`
```
## Subject

The `FakeTemplating` class is used across several bundles and should be shared in the public `Test` namespace.

### SonataDashboardBundle
Tests/Block/ContainerBlockServiceTest.php:use Sonata\BlockBundle\Tests\Block\Service\FakeTemplating;

### SonataMediaBundle
Tests/Block/GalleryListBlockServiceTest.php:use Sonata\BlockBundle\Tests\Block\Service\FakeTemplating;

### SonataPageBundle
Tests/Block/PageListBlockServiceTest.php:use Sonata\BlockBundle\Tests\Block\Service\FakeTemplating;

### SonataSeoBundle
Tests/Block/Social/FacebookLikeBoxBlockServiceTest.php:use Sonata\BlockBundle\Tests\Block\Service\FakeTemplating;
Tests/Block/Social/FacebookLikeButtonBlockServiceTest.php:use Sonata\BlockBundle\Tests\Block\Service\FakeTemplating;
Tests/Block/Social/FacebookSendButtonBlockServiceTest.php:use Sonata\BlockBundle\Tests\Block\Service\FakeTemplating;
Tests/Block/Social/FacebookShareButtonBlockServiceTest.php:use Sonata\BlockBundle\Tests\Block\Service\FakeTemplating;

### DIFF
--- a/Test/FakeTemplating.php
+++ b/Test/FakeTemplating.php
@@ -1,0 +1,84 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\BlockBundle\Test;
+
+use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Mocking class for template usage.
+ *
+ * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ */
+class FakeTemplating implements EngineInterface
+{
+    /**
+     * @var string
+     */
+    public $view;
+
+    /**
+     * @var array
+     */
+    public $parameters;
+
+    /**
+     * @var Response
+     */
+    public $response;
+
+    /**
+     * @var string
+     */
+    public $name;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function render($name, array $parameters = array())
+    {
+        $this->name = $name;
+        $this->parameters = $parameters;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function renderResponse($view, array $parameters = array(), Response $response = null)
+    {
+        $this->view = $view;
+        $this->parameters = $parameters;
+        $this->response = $response;
+
+        if ($response) {
+            return $response;
+        }
+
+        return new Response();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supports($name)
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function exists($name)
+    {
+        return true;
+    }
+}
--- a/Tests/Block/Service/FakeTemplating.php
+++ b/Tests/Block/Service/FakeTemplating.php
@@ -11,68 +11,17 @@
 
 namespace Sonata\BlockBundle\Tests\Block\Service;
 
-use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
-use Symfony\Component\HttpFoundation\Response;
+use Sonata\BlockBundle\Test\FakeTemplating as BaseTemplating;
 
-class FakeTemplating implements EngineInterface
+@trigger_error(
+    'The '.__NAMESPACE__.'\FakeTemplating class is deprecated since version 3.x and will be removed in 4.0.'
+    .' Use Sonata\BlockBundle\Test\FakeTemplating instead.',
+    E_USER_DEPRECATED
+);
+
+/**
+ * @deprecated since version 3.x and will be removed in 4.0. Use Sonata\BlockBundle\Test\FakeTemplating instead.
+ */
+class FakeTemplating extends BaseTemplating
 {
-    /**
-     * @var string
-     */
-    public $view;
-
-    /**
-     * @var array
-     */
-    public $parameters;
-
-    /**
-     * @var Response
-     */
-    public $response;
-
-    /**
-     * @var string
-     */
-    public $name;
-
-    /**
-     * {@inheritdoc}
-     */
-    public function render($name, array $parameters = array())
-    {
-        $this->name = $name;
-        $this->parameters = $parameters;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function renderResponse($view, array $parameters = array(), Response $response = null)
-    {
-        $this->view = $view;
-        $this->parameters = $parameters;
-
-        if ($response) {
-            return $response;
-        }
-
-        return new Response();
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function supports($name)
-    {
-        return true;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function exists($name)
-    {
-        return true;
-    }
 }
--- a/Test/AbstractBlockServiceTestCase.php
+++ b/Test/AbstractBlockServiceTestCase.php
@@ -16,7 +16,6 @@ use Sonata\BlockBundle\Block\BlockContextManager;
 use Sonata\BlockBundle\Block\BlockContextManagerInterface;
 use Sonata\BlockBundle\Block\BlockServiceInterface;
 use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
-use Sonata\BlockBundle\Tests\Block\Service\FakeTemplating;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
--- a/Tests/Test/FakeTemplatingTest.php
+++ b/Tests/Test/FakeTemplatingTest.php
@@ -1,0 +1,61 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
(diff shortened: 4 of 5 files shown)


core23:
Not sure if this is the correct namespace? Maybe `Test\Fixtures`?

greg0ire:
What about `Test\Mock` ?

Test/FakeTemplating.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+namespace Sonata\BlockBundle\Test;
+
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Templating\EngineInterface;
+
+/**
+ * @author Christian Gripp <mail@core23.de>
```
Soullivaneuh: @rande is the author actually. :stuck_out_tongue:

Soullivaneuh:
> Not sure if this is the correct namespace? Maybe `Test\Fixtures`?

It depends of how this is used IMHO.

BTW, if you move it to the public namespace, it should:

* Have a usage description on the phpdoc.
* Be documented.

core23:
Renamed the `FakeTemplating` class to `MockTemplating`, because it is a mock :trollface: 

Do we need a docs for a mock, because all methods are defined in the interface @Soullivaneuh ?

There are some tests who are using the mock.

Soullivaneuh:
> Renamed the `FakeTemplating` class to `MockTemplating`, because it is a mock :trollface: 

`TemplatingMock` maybe?

> Do we need a docs for a mock, because all methods are defined in the interface @Soullivaneuh ?

A usage example yes.

core23:
> `TemplatingMock` maybe?

Seen both names in symfony, `Mock` infront and at the end :/

core23:
> A usage example yes.

As I said, we have tests who are using this mock. This is just a refactoring of the name(space).

core23:
Can you please have a look @Soullivaneuh ?

We need this to remove the test dependencies in the other bundles. IMHO we don't need any further information for a mocking class.

Test/Mock/MockTemplating.php
```diff
@@ -0,0 +1,83 @@
...
+
+use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Mocking class for template usage.
+ *
+ * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ */
+class MockTemplating implements EngineInterface
```
Soullivaneuh: `TemplatingMock` would be better.

If we follow your naming, all class would me `Mock*`

It's a suffix for interfaces and trait. I think this should be the same here.

cc @greg0ire @rande 

Also, maybe this class can be final?

Soullivaneuh:
> As I said, we have tests who are using this mock. This is just a refactoring of the name(space).

This is not a question of "we have test" or not. This classes is officially moved on the public namespace, for end user.

Every public API should be tested and documented. I don't see why this should be an exception. :wink:

Test/Mock/MockTemplating.php
core23: But it's a prefix for abstracts...

core23: We can't make it final now, because the old `FakeTemplating` class is extending this class for BC.

greg0ire:
I read the draft of a blog post from @roukmoute today and learnt that dummies, mocks, spies and fakes are all test doubles. How about using this as a namespace?

core23:
Can you please elaborate what should be done here? Which namespace do you prefer?

Test/Mock/MockTemplating.php
Soullivaneuh: > But it's a prefix for abstracts...

Indeed... I don't know. You said Symfony does both? Oo

core23: **Suffix**
- Symfony\Bridge\PhpUnit\ClockMock
- Symfony\Bridge\PhpUnit\DnsMock


**Prefix**
- Symfony\Component\Finder\Tests\Iterator\MockSplFileInfo
- Symfony\Component\Filesystem\Tests\Fixtures\MockStream\MockStream

greg0ire:
`Test\Double` feels great for test doubles, WDYT?

core23:
> `Test\Double` feels great for test doubles, WDYT?

This is no symfony standard. Haven't found any package who uses this structure.

greg0ire:
Does symfony have a standard for this?

core23:
No, as I said they have several namesspaces with different class names.

greg0ire:
What I'm saying is since dummies, fakes, spies and mocks are all different things, and are all test doubles, it would make much sense to file them all under the same namespace, wouldn't it ?

- `Test\Double\FooDummy`
- `Test\Double\BarFake`
- `Test\Double\FoobarSpy`
- `Test\Double\FizzBuzzMock`

What do @sonata-project/contributors think about this organization ?

OskarStark:
Sounds good to me 👍

Soullivaneuh:
Not sure it will worth.

Classes like `BarFake` are generally for fixtures.

Even if we put all of them onto `Test\Double`, we should separate the namespace for `Mock` IMHO.

greg0ire:
For me fixtures are yml, or xml files that contain data, as opposed to behavior. I would put that in a separate directory.

Soullivaneuh:
> For me fixtures are yml, or xml files that contain data

A fixtures is not necessary about data. :-)

greg0ire:
Hence the name of [this library](https://github.com/doctrine/data-fixtures) I suppose… you're right it seems. So I'm ok for `Fixtures` as the top namespace, although I believe `Double` would be a bit less confusing.

Soullivaneuh:
In which case we would ave another think than a Mock class?

In this case it will be particular and should be outside the scope.

So `Test\Mock` or if you really want double, `Test\Double\Mock`. But I think it would be a bit cumbersome, no?

greg0ire:
> In which case we would ave another think than a Mock class?

If we have a fake, a dummy, or a spy, which are different from a mock, I'm told. It seems that there is a lot of confusion in this area, and it would be great to get that right.

BTW, @roukmoute, could you tell us what the templating class here is? I think it actually is a stub, because it does not match the definition of a mock you give since it is not the "complete substitution of a concrete class + some code that checks the behavior", but rather "a tiny implementation of the interface", so… a fake.

roukmoute:
Hey hey :smile: !

I think we are closer to a Fake, borderline with a Stub.
A Mock is always a Spy.
But a Mock check his own **behavior verification**, so a Mock can fail.

A Stub/Fake must succeed because it performs a specific test, Mock can, for example, if he has no good arguments for an method, decide to fail.

greg0ire:
Thanks for the explanation @roukmoute! 

> Even if we put all of them onto Test\Double, we should separate the namespace for Mock IMHO.

Why? Isn't the `Mock` prefix / suffix enough ? Plus if we have 1 fake, 1 stub, 2 mocks and 1 spy, won't it be cumbersome to create one directory for each type ? People will probably often confuse mocks and spies, etc… (that is, until they read @roukmoute's article), so we must be ready to quickly rename these classes. @roukmoute also tells me that doubles are fixtures, which sounds like what you are telling me @Soullivaneuh , and he suggests to use `Test\Fixtures\Double\FooMock` (and probably `Test\Fixtures\data\user.yml` if we have data fixtures)

Soullivaneuh:
After re-reading I agree with @roukmoute. It's a fake, not a mock.

A mock is internal to PHPUnit. It could be a mock if the class extends the mock class, but this is not the case here (and I don't know if it's even possible).

core23:
Rerenamed the class to `FakeTemplating`.

Tests/Block/Service/FakeTemplating.php
```diff
@@ -11,68 +11,17 @@
...
-
-    /**
-     * {@inheritdoc}
-     */
-    public function exists($name)
-    {
-        return true;
-    }
+/**
+ * @deprecated Deprecated since version 3.x. Use Sonata\BlockBundle\Test\FakeTemplating instead.
```
Soullivaneuh: To be removed in 4.0.

core23: Fixed

Soullivaneuh:
Code looks good! :+1: 

Missing the documentation to get this complete.

https://github.com/sonata-project/SonataBlockBundle/pull/310#issuecomment-230001570

core23:
Not sure what kind of docu are you expecting... We have no new PHPdoc, because we are faking an interface implementation. and we have existing tests who are using the interface. Also, we don't have any kind of RST documentation for any other test case.

Soullivaneuh:
> Also, we don't have any kind of RST documentation for any other test case.

Because we never made real public test case. Time to start. :wink:

Soullivaneuh:
For the doc: Maybe a simple test case usage and a description of how to use it and when?

Soullivaneuh:
@core23 Could you please rebase?

I'll push it on 3.1.1 for now. Could you please work on a little documentation after? Thanks!

core23:
Rebased @Soullivaneuh 

I would start working on a documentation afterwards.

Soullivaneuh:
@core23 https://github.com/sonata-project/SonataBlockBundle/releases/tag/3.1.1

PR #317, opened 2016-07-10
Fixed cs

core23:
I am targetting this branch, because this is a cs fix only.

## Subject

Fixed some cs.


OskarStark:
Thank you @core23

PR #318, opened 2016-07-10
Remove BaseBlockService

core23:
I am targetting this branch, because this can't be done completly without a BC break.

## Changelog

```markdown
### Added
- Added `AbstractAdminBlockService`

### Changed
- Moved general block methods to `AbstractBlockService`

### Removed
- Removed `BaseBlockService`
```

## To do

- [X] Update the tests
- [ ] Add an upgrade note

## Subject

The class `BaseBlockService` doesn't match the correct naming convention and starts with `Base` instead of `Abstract`. The `BaseBlockService` class contains special methods for the admin bundle, so this were moved to the new `AbstractAdminBlockService` class.

### BC Break

Because the class `AbstractBlockService` exists, we can't move all methods and the constructor to abstract class without a break. Our bundles don't use the class `AbstractBlockService` at the moment, but someone else could do it.


core23:
Please leave your feedback to this change. If you're okay with it, I would try to create a PR for the stable branch with some deprecation warnings.

core23:
What dou you think @OskarStark @greg0ire @Soullivaneuh ? Should we remove the `BaseBlockService` and create a new `AbstractAdminBlockService`?

greg0ire:
What about having a `Sonata\BlockBundle\Bridge\AdminBundle\Block` namespace instead? Wouldn't it make things a bit clearer?

core23:
> What about having a `Sonata\BlockBundle\Bridge\AdminBundle\Block` namespace instead? Wouldn't it make things a bit clearer?

IMO, this not a really a bridge, because it's no abstraction of a concreate implementation.

core23:
Maybe move the `AbstractAdminBlockService` to the admin bundle instead?

rande:
Cannot we have a BC layer ? This might break a lot of block in other bundles

core23:
Closing this in favor of https://github.com/sonata-project/SonataBlockBundle/pull/328

PR #319, opened 2016-07-11
Redesign Block profiler

qsomazzi:
## Changelog

```markdown
### Fixed
- Profiler block design for Symfony Profiler v2
```

## Subject

This is just a redesign for Blocks profiler. Its was not readable with SF3.

BEFORE : 
before

AFTER:
capture d ecran 2016-07-11 a 16 49 05

### DIFF
--- a/Resources/views/Profiler/block.html.twig
+++ b/Resources/views/Profiler/block.html.twig
@@ -59,6 +59,9 @@
 {% endblock %}
 
 {% block panel %}
+    {# NEXT_MAJOR : remove this when Symfony Requirement will be bumped to 2.8+ #}
+    {% set profiler_markup_version = profiler_markup_version|default(1) %}
+
     <h2>Events Blocks</h2>
     <table>
         <tr>
@@ -92,11 +95,23 @@
 
     <h2>Real Blocks</h2>
     {% set blocks = collector.realBlocks %}
-    {{ block('table') }}
+    {% if profiler_markup_version == 1 %}
+        {{ block('table') }}
+    {% else %}
+        <div class="tab-content">
+            {{ block('table_v2') }}
+        </div>
+    {% endif %}
 
     <h2>Containers Blocks</h2>
     {% set blocks = collector.containers %}
-    {{ block('table') }}
+    {% if profiler_markup_version == 1 %}
+        {{ block('table') }}
+    {% else %}
+        <div class="tab-content">
+            {{ block('table_v2') }}
+        </div>
+    {% endif %}
 
     {{ block('javascript') }}
 {% endblock %}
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
Did you check the compatibility with the old profiler? Can we have screenshots about this?

rande:
I think we need to add switch case depends on the symfony version.

qsomazzi:
in fact, I haven't checked it, because for me, the goal of this is to handle new profiler, not to be compatible with the old one.

qsomazzi:
@Soullivaneuh @rande symfony version check added

DependencyInjection/Configuration.php
```diff
@@ -71,7 +72,7 @@ public function getConfigTreeBuilder()
                     ->fixXmlConfig('container_type', 'container_types')
                     ->children()
                         ->scalarNode('enabled')->defaultValue('%kernel.debug%')->end()
-                        ->scalarNode('template')->defaultValue('SonataBlockBundle:Profiler:block.html.twig')->end()
+                        ->scalarNode('template')->defaultValue(version_compare(Kernel::VERSION, '2.7', '<') ? 'SonataBlockBundle:Profiler:block.html.twig' : 'SonataBlockBundle:Profiler:block_v3.html.twig')->end()
```
OskarStark: The Kernel Version is NOT the Symfony Version!

Tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -36,7 +37,7 @@ public function testOptions()
             ),
             'profiler' => array(
                 'enabled' => '%kernel.debug%',
-                'template' => 'SonataBlockBundle:Profiler:block.html.twig',
+                'template' => version_compare(Kernel::VERSION, '2.7', '<') ? 'SonataBlockBundle:Profiler:block.html.twig' : 'SonataBlockBundle:Profiler:block_v3.html.twig',
```
OskarStark: same here

DependencyInjection/Configuration.php
qsomazzi: how can I check this ? I can't find another way to get it

Soullivaneuh: You can use the `profiler_markup_version` variable directly on Twig.

More info: https://github.com/symfony/symfony/blob/2.8/UPGRADE-2.8.md#webprofiler

qsomazzi:
thanks @Soullivaneuh, corrected

Resources/views/Profiler/block.html.twig
```diff
@@ -157,3 +171,63 @@
         {% endfor %}
     </table>
 {% endblock %}
+
+{% block tablev3 %}
```
Soullivaneuh: Should be v2 IMHO, because it's the v2 of the profiler. :wink:

Soullivaneuh: And with an underscore please. :-)

Soullivaneuh:
Could we have a screenshot of the old profiler too?

Resources/views/Profiler/block.html.twig
```diff
@@ -59,6 +59,8 @@
 {% endblock %}
 
 {% block panel %}
+    {% set profiler_markup_version = profiler_markup_version|default(1) %}
```
Soullivaneuh: Please add a `NEXT_MAJOR` comment telling this can be removed when Symfony requirement will be bumped to 2.8+.

qsomazzi:
@Soullivaneuh I didn't change anything on the old profiler, and i don't have it installed on my current project, can we do without it ?

Soullivaneuh:
Indeed. Ok for this.

LGTM. cc @sonata-project/contributors for another review.

Soullivaneuh:
Oh and also please add a changelog on the PR subject.

qsomazzi:
Changelog added

Soullivaneuh:
Ok so first, the markdown block must be kept.

Second, a changelog is not just listing the modified file, we have git for that. :wink: 

I updated yours. Please take a look of the current one: https://github.com/sonata-project/SonataBlockBundle/blob/3.x/CHANGELOG.md

qsomazzi:
Thanks, now that I understand how to do this, I will do it better next time :)

rande:
Ok for me!

qsomazzi:
@Soullivaneuh good to merge ? :)

OskarStark:
LGTM 👍

Soullivaneuh:
Thank you @qsomazzi! :+1:

qsomazzi:
Great :) thanks for your help !

PR #321, opened 2016-07-14
Removed ugly symfony2 workaround

### LINKED ISSUE #313: Remove ugly symfony2 workaround
## Subject

Remove the ugly workaround in `composer.json`:

```json
    "autoload": {
        "files": ["Resources/stubs/symfony2.php"]
    },
```

Don't know why we have this workaround, but we should remove this or put it somewhere in the code.

core23:
I am targetting this branch, because this is a very ugly fix for newer symfony versions.

Closes #313

## Changelog

```markdown
### Removed
 - Removed ugly workaround for `OptionsResolverInterface`
```

## Subject

Removed pretty ugly symfony BC helper.


Soullivaneuh:
Like this one but are you sure it will not throw any edge case?

core23:
Not as easy, as I thought... Think we can fix this with the next major release, when dropping `BlockContextManager::setDefaultSettings`.

Soullivaneuh:
@core23 What about removed the class type from the signature?

Now:

```php
protected function setDefaultSettings(OptionsResolverInterface $optionsResolver, BlockInterface $block)
```

Then:

```php
protected function setDefaultSettings($optionsResolver, BlockInterface $block)
```

And add a `instanceof` condition instead?

core23:
Good point. I'll check this later

core23:
We can't do this. This is a BC break too https://3v4l.org/niGD0

PR #328, opened 2016-07-29
Deprecate BaseBlockService and BlockAdminServiceInterface

core23:
I am targetting this branch, because this is BC.

## Changelog

```markdown
### Added
- created `Sonata\BlockBundle\Block\Service\AbstractAdminBlockService` class
- created `Sonata\BlockBundle\Block\Service\AbstractBlockService` class
- created `Sonata\BlockBundle\Block\Service\AdminBlockServiceInterface` class
- created `Sonata\BlockBundle\Block\Service\BlockServiceInterface` class

### Deprecated
- the class `Sonata\BlockBundle\Block\AbstractBlockService` is deprecated
- the class `Sonata\BlockBundle\Block\BaseBlockService` is deprecated
- the class `Sonata\BlockBundle\Block\BlockAdminServiceInterface` is deprecated
- the class `Sonata\BlockBundle\Block\BlockServiceInterface` is deprecated

```

## To do

- [X] Update the tests
- [x] Update the documentation
- [X] Add an upgrade note

## Subject

The class BaseBlockService doesn't match the correct naming convention and starts with Base instead of Abstract. The BaseBlockService class contains special methods for the admin bundle, so should be moved to special class.

PR #330, opened 2016-08-12
Revert "Deprecate BaseBlockService and BlockAdminServiceInterface"

core23:
Reverts sonata-project/SonataBlockBundle#328

We can't use this right now, because ``BlockRenderer::addMetaInformation`` will break with this.


OskarStark:
Uhh, sorry for this @core23 , didn't get it 😕

PR #344, opened 2016-11-04
Always use service id as block id

core23:
This is the BC version for #302

### Changelog

``` markdown
### Changed
- empty block names are automatically set via `TweakCompilerPass`
```
### Subject

The block id is automatically set to the internal service id (if empty). This is the same mechanism like creating a admin.

### DIFF
--- a/DependencyInjection/Compiler/TweakCompilerPass.php
+++ b/DependencyInjection/Compiler/TweakCompilerPass.php
@@ -31,12 +31,27 @@ class TweakCompilerPass implements CompilerPassInterface
 
         $parameters = $container->getParameter('sonata_block.blocks');
 
-        foreach ($container->findTaggedServiceIds('sonata.block') as $id => $attributes) {
+        foreach ($container->findTaggedServiceIds('sonata.block') as $id => $tags) {
+            $definition = $container->getDefinition($id);
+
+            $arguments = $definition->getArguments();
+
+            // Replace empty block id with service id
+            if (strlen($arguments[0]) == 0) {
+                $definition->replaceArgument(0, $id);
+            } elseif ($id != $arguments[0]) {
+                // NEXT_MAJOR: Remove deprecation notice
+                @trigger_error(
+                    'Using different service id and block id is deprecated since 3.x and will be removed in 4.0.',
+                    E_USER_DEPRECATED
+                );
+            }
+
             $manager->addMethodCall('add', array($id, $id, isset($parameters[$id]) ? $parameters[$id]['contexts'] : array()));
         }
 
         $services = array();
-        foreach ($container->findTaggedServiceIds('sonata.block.loader') as $id => $attributes) {
+        foreach ($container->findTaggedServiceIds('sonata.block.loader') as $id => $tags) {
             $services[] = new Reference($id);
         }
 
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -1,6 +1,26 @@
 UPGRADE 3.x
 ===========
 
+Injecting the block id into a service is deprecated and will be automatically set.
+
+Instead, provide an empty argument:
+
+## Before
+```
+    <service id="acme.block.service" class="Acme\BlockBundle\AcmeBlockService">
+        <tag name="sonata.block"/>
+        <argument>acme.block.service</argument>
+    </service>
+```
+
+## After
+```
+    <service id="acme.block.service" class="Acme\BlockBundle\AcmeBlockService">
+        <tag name="sonata.block"/>
+        <argument/>
+    </service>
+```
+
 UPGRADE FROM 3.1 to 3.2
 =======================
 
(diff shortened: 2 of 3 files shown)


greg0ire:
Why do we want that again? I read #302, but it's unclear to me.

DependencyInjection/Compiler/TweakCompilerPass.php
```diff
@@ -31,12 +31,31 @@ public function process(ContainerBuilder $container)
...
+                // NEXT_MAJOR: Remove deprecation notice
+                @trigger_error(
+                    'Using different service id and block id is deprecated since 3.x.',
+                    E_USER_DEPRECATED
+                );
+            }
+
+            if (strlen($arguments[0]) == 0) {
+                $definition->replaceArgument(0, $id);
+            }
```
greg0ire: Uuuuuh… again :open_mouth: ?

DependencyInjection/Compiler/TweakCompilerPass.php
```diff
@@ -31,12 +31,31 @@ public function process(ContainerBuilder $container)
 
         $parameters = $container->getParameter('sonata_block.blocks');
 
-        foreach ($container->findTaggedServiceIds('sonata.block') as $id => $attributes) {
+        foreach ($container->findTaggedServiceIds('sonata.block') as $id => $tags) {
```
greg0ire: `$tags` is not used, you could write `array_keys($container->findTaggedServiceIds('sonata.block'))` instead, might be less confusing

UPGRADE-3.x.md
```diff
@@ -1,6 +1,8 @@
 UPGRADE 3.x
 ===========
 
+Injecting the block id is deprecated and will be automatically set.
```
greg0ire: What should the user do though? Leave an empty `<argument/>` ?

If yes please add :

```
Instead, provide an empty (null?) argument:

## Before
`<argument>your.block</argument>`

## After
`<argument/>`
```

DependencyInjection/Compiler/TweakCompilerPass.php
core23: Working with `array_keys` here looks more confusing to me

DependencyInjection/Compiler/TweakCompilerPass.php
core23: Copy & paste problem ;)

core23:
> Why do we want that again? I read #302, but it's unclear to me.

ATM it's a little confusing, that we have two different ids for a `BlockService`, the service id and the block id. We should change that part, so that the id's are corresponding. We use the same mechanism for [admin services ](https://github.com/sonata-project/SonataAdminBundle/blob/3.x/DependencyInjection/Compiler/AddDependencyCallsCompilerPass.php#L50-L58)(admin id = service id).

DependencyInjection/Compiler/TweakCompilerPass.php
greg0ire: Ok, it must be a matter of taste, so do as you wish. Be aware that `phpmd` will whine about this kind of thing though.

greg0ire:
Try amend + force push, for StyleCI?

DependencyInjection/Compiler/TweakCompilerPass.php
core23: That's not the only code where `phpmd` will whine ;)

greg0ire:
:confused: Maybe contact their support (there is an email on the StyleCI page)?

greg0ire:
Can you rebase @core23 ? I fixed the Travis build : https://github.com/sonata-project/SonataBlockBundle/pull/345/files#diff-0e4afb0bb3966e755707777f8ec7f823L4

core23:
Rebased

greg0ire:
@sonata-project/contributors please review

DependencyInjection/Compiler/TweakCompilerPass.php
```diff
@@ -31,12 +31,27 @@ public function process(ContainerBuilder $container)
...
+
+            $arguments = $definition->getArguments();
+
+            // Replace empty block id with service id
+            if (strlen($arguments[0]) == 0) {
+                $definition->replaceArgument(0, $id);
+            } elseif ($id != $arguments[0]) {
+                // NEXT_MAJOR: Remove deprecation notice
+                @trigger_error(
+                    'Using different service id and block id is deprecated since 3.x.',
```
OskarStark: And will be removed in 4.0

core23: Fixed that too

OskarStark:
Thank you @core23 👍

Jean85:
This BC/deprecation is missing from the changelog and from the upgrade guide.
I would write a PR myself, but I'm not understanding how to fix this...

core23:
That's a known problem, because we can't rename this services without a BC break.

PR #357, opened 2017-01-22
Removed deprecated getMock in unit tests

core23:
I am targetting this branch, because this is for tests only.

## Changelog

```markdown
### Changed
 - `AbstractBlockServiceTestCase` uses `createMock` method
```

## Subject

The method `getMock` is deprecated.

### DIFF
--- a/Test/AbstractBlockServiceTestCase.php
+++ b/Test/AbstractBlockServiceTestCase.php
@@ -47,11 +47,11 @@ abstract class AbstractBlockServiceTestCase extends \PHPUnit_Framework_TestCase
 
     protected function setUp()
     {
-        $this->container = $this->getMock('Symfony\Component\DependencyInjection\ContainerInterface');
+        $this->container = $this->createMock('Symfony\Component\DependencyInjection\ContainerInterface');
         $this->templating = new FakeTemplating();
 
-        $blockLoader = $this->getMock('Sonata\BlockBundle\Block\BlockLoaderInterface');
-        $this->blockServiceManager = $this->getMock('Sonata\BlockBundle\Block\BlockServiceManagerInterface');
+        $blockLoader = $this->createMock('Sonata\BlockBundle\Block\BlockLoaderInterface');
+        $this->blockServiceManager = $this->createMock('Sonata\BlockBundle\Block\BlockServiceManagerInterface');
         $this->blockContextManager = new BlockContextManager($blockLoader, $this->blockServiceManager);
     }
 
@@ -66,7 +66,7 @@ abstract class AbstractBlockServiceTestCase extends \PHPUnit_Framework_TestCase
     {
         $this->blockServiceManager->expects($this->once())->method('get')->will($this->returnValue($blockService));
 
-        $block = $this->getMock('Sonata\BlockBundle\Model\BlockInterface');
+        $block = $this->createMock('Sonata\BlockBundle\Model\BlockInterface');
         $block->expects($this->once())->method('getSettings')->will($this->returnValue(array()));
 
         $blockContext = $this->blockContextManager->get($block);
@@ -96,5 +96,21 @@ abstract class AbstractBlockServiceTestCase extends \PHPUnit_Framework_TestCase
         ksort($blockSettings);
 
         $this->assertSame($completeExpectedOptions, $blockSettings);
+    }
+
+    /**
+     * NEXT_MAJOR: Remove this method when dropping support for < PHPUnit 5.4.
+     *
+     * @param string $class
+     *
+     * @return \PHPUnit_Framework_MockObject_MockObject
+     */
+    protected function createMock($class)
+    {
+        if (is_callable('parent::createMock')) {
+            return parent::createMock($class);
+        }
+
+        return $this->getMock($class);
     }
 }
--- a/Tests/Block/BlockContextManagerTest.php
+++ b/Tests/Block/BlockContextManagerTest.php
@@ -13,21 +13,22 @@ namespace Sonata\BlockBundle\Tests\Block;
 
 use Doctrine\Common\Util\ClassUtils;
 use Sonata\BlockBundle\Block\BlockContextManager;
+use Sonata\BlockBundle\Tests\PHPUnit_Framework_TestCase;
 
-class BlockContextManagerTest extends \PHPUnit_Framework_TestCase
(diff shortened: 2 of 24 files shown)


core23:
This is one more reason to [include the phpunit dependency](https://github.com/sonata-project/dev-kit/issues/197)!

cc @sonata-project/contributors

greg0ire:
This cannot work with php < 5.6

core23:
You're right :( But the phpunit problem itself still exists. Nobody knows which version do we use in our tests.

core23:
Create a simple method for external test classes, so they would run on old and new PHPunit versions.

greg0ire:
I'm 👍 with this, please do it for all tests.

core23:
Not sure how to do this. Should we create a new static wrapper class for our internal tests or just copy&paste the same method several times?

greg0ire:
The static class seems less intrusive, I'd use that. It's ugly, but bound to disappear, so, not that big of a deal.

osavchenko:
Nope, static class is not good idea: `Non-static method PHPUnit_Framework_TestCase::createMock() should not be called statically`

greg0ire:
Uh you're right. We don't have traits yet so yeah, wrapper class.

Tests/Block/BlockContextManagerTest.php
```diff
@@ -114,15 +114,15 @@ public function testWithInvalidSettings()
...
 //
-//        $blockLoader = $this->getMock('Sonata\BlockBundle\Block\BlockLoaderInterface');
+//        $blockLoader = $this->createMock('Sonata\BlockBundle\Block\BlockLoaderInterface');
 //
-//        $serviceManager = $this->getMock('Sonata\BlockBundle\Block\BlockServiceManagerInterface');
+//        $serviceManager = $this->createMock('Sonata\BlockBundle\Block\BlockServiceManagerInterface');
 //        $serviceManager->expects($this->exactly(2))->method('get')->will($this->returnValue($service));
 //
-//        $block = $this->getMock('Sonata\BlockBundle\Model\BlockInterface');
+//        $block = $this->createMock('Sonata\BlockBundle\Model\BlockInterface');
```
osavchenko: Why you comment this code?

Test/AbstractBlockServiceTestCase.php
```diff
@@ -97,4 +97,20 @@ protected function assertSettings(array $expected, BlockContextInterface $blockC
...
     }
+
+    /**
+     * NEXT_MAJOR: Remove this method when dropping support for < PHPUnit 5.4.
+     *
+     * @param string $class
+     *
+     * @return \PHPUnit_Framework_MockObject_MockObject
+     */
+    protected function createMock($class)
```
osavchenko: What about some base class to avoid code duplication? You can create something like [this](https://github.com/sonata-project/SonataAdminBundle/pull/4288/commits/6ee98abefe3f791e23669fc8c2de3033c50d3769) and avoid this method in each test class.

Tests/Block/BlockContextManagerTest.php
core23: Was you can see, I didn't comment this method I just fixed the call. Also we could remove the dead code...

Test/AbstractBlockServiceTestCase.php
core23: Not sure about the `right` way. Both have it pros and cons.

Tests/Block/BlockContextManagerTest.php
osavchenko: oh, yes! Sorry for this.

Test/AbstractBlockServiceTestCase.php
osavchenko: But removing a lot of _same_ methods will be annoying. Isn't it?

greg0ire: > Not sure about the right way.

No need to be ultra-right here, it's code that will disappear, so just do what you feel is simpler, bearing in mind that if we go @osavchenko , we will still have to change the `extends` one way, and the other way around when removing this on next major.

Soullivaneuh: > we will still have to change the `extends`

Maybe, but it's just a line.

If we don't create a base class, in this case, it would be the same to add a condition to each `createMock` calls instead.

BTW, this is the way we are doing for SF BC, why not here?

greg0ire:
Is the Travis failure related to this PR?

osavchenko:
No, @greg0ire. It's bug for old phpunit versions.

Tests/Block/Service/RssBlockServiceTest.php
```diff
@@ -37,7 +37,7 @@ public function testService()
 
         $blockContext = new BlockContext($block, $optionResolver->resolve());
 
-        $formMapper = $this->getMock('Sonata\\AdminBundle\\Form\\FormMapper', array(), array(), '', false);
+        $formMapper = $this->createMock('Sonata\\AdminBundle\\Form\\FormMapper', array(), array(), '', false);
```
osavchenko: Should be:

```php
$formMapper = $this->getMockBuilder('Sonata\\AdminBundle\\Form\\FormMapper')
            ->disableOriginalConstructor()
            ->getMock();
```

Tests/Block/Service/TextBlockServiceTest.php
```diff
@@ -34,7 +34,7 @@ public function testService()
 
         $blockContext = new BlockContext($block, $optionResolver->resolve($block->getSettings()));
 
-        $formMapper = $this->getMock('Sonata\\AdminBundle\\Form\\FormMapper', array(), array(), '', false);
+        $formMapper = $this->createMock('Sonata\\AdminBundle\\Form\\FormMapper', array(), array(), '', false);
```
osavchenko: Should be

```php
$formMapper = $this->getMockBuilder('Sonata\\AdminBundle\\Form\\FormMapper')
            ->disableOriginalConstructor()
            ->getMock();
```

Tests/Block/Service/RssBlockServiceTest.php
core23: Thanks @osavchenko 👍

greg0ire:
Thanks @core23 !

Issue #363, opened 2017-02-02
3.3.0 throwing strange deprecation notices


mmucklo:
Seeing some strange deprecation notices since updating

### Environment

dev (eng)

#### Sonata packages

```bash
 sonata-project/admin-bundle                  3.12.0            
  sonata-project/block-bundle                  3.3.0             
  sonata-project/cache                         1.0.7             
  sonata-project/core-bundle                   3.2.0             
  sonata-project/datagrid-bundle               2.2               
  sonata-project/doctrine-extensions           dev-master fbaff00
  sonata-project/doctrine-orm-admin-bundle     dev-master ab4ba73
  sonata-project/easy-extends-bundle           dev-master 36bac7e
  sonata-project/exporter                      1.7.0         
```

#### Symfony packages

```bash
  symfony/assetic-bundle                       dev-master 0241b13
  symfony/monolog-bundle                       v3.0.3            
  symfony/phpunit-bridge                       v3.2.2            
  symfony/polyfill-intl-icu                    v1.3.0            
  symfony/polyfill-mbstring                    v1.3.0            
  symfony/polyfill-php54                       v1.3.0            
  symfony/polyfill-php56                       v1.3.0            
  symfony/polyfill-php70                       v1.3.0            
  symfony/polyfill-util                        v1.3.0            
  symfony/psr-http-message-bridge              v0.3              
  symfony/security-acl                         v3.0.0            
  symfony/swiftmailer-bundle                   v2.4.2            
```

#### PHP version

```bash
$ php -v
PHP 7.0.15 (cli) (built: Feb  1 2017 01:00:04) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Xdebug v2.5.0, Copyright (c) 2002-2016, by Derick Rethans
    with Zend OPcache v7.0.15, Copyright (c) 1999-2017, by Zend Technologies```
```

## Subject
See the deprecation notices I'm seeing.

I'm not using any custom blocks at this point.

## Steps to reproduce
run phpunit on a clean cache

## Expected results
no sonata deprecation notices

## Actual results
```
Using different service id and block id is deprecated since 3.x and will be removed in 4.0: 6x
```

Coming from file: vendor/sonata-project/block-bundle/DependencyInjection/Compiler/TweakCompilerPass.php

```php
            if (strlen($arguments[0]) == 0) {
[...]

PR #366, opened 2017-02-27
improve deprecation notice

dbu:
I am targetting this branch, because the deprecation notice is in the 3.x branch.

## Subject

It is really hard to know what is wrong when the deprecation message just says "Using different service id and block id is deprecated". Even when looking at a stack trace, we only see that it happens during symfony building the container, but not where it happened. With the id's it becomes trivial to find the place were the mistake happens.

I did not add a changelog because this fix on the deprecation notice seems not something that should appear in the changelog.

(Btw, i stumbled upon this while investigating deprecations - turns out the problem is within the block bundle itself. It would be nice if we could supress the deprecation notice in that case - i assume its not changed to preserve BC. we could also possibly move the services and create an alias from the old service. Unless somebody did a compiler pass to change the service definition...)

### DIFF
--- a/DependencyInjection/Compiler/TweakCompilerPass.php
+++ b/DependencyInjection/Compiler/TweakCompilerPass.php
@@ -43,7 +43,7 @@ class TweakCompilerPass implements CompilerPassInterface
             } elseif ($id != $arguments[0]) {
                 // NEXT_MAJOR: Remove deprecation notice
                 @trigger_error(
-                    'Using different service id and block id is deprecated since 3.x and will be removed in 4.0.',
+                    sprintf('Using service id %s different from block id %s is deprecated since 3.x and will be removed in 4.0.', $id, $arguments[0]),
                     E_USER_DEPRECATED
                 );
             }


greg0ire:
Labelling as pedantic since it's not deemed important enough to warrant a changelog entry.

greg0ire:
> turns out the problem is within the block bundle itself. It would be nice if we could supress the deprecation notice in that case - i assume its not changed to preserve BC.

@core23 , can you help about this?

dbu:
> pedantic

exactly :-)

> supress deprecation / move service and alias

=> this should not block the merge request. shall i open an issue about that?

greg0ire:
> this should not block the merge request. shall i open an issue about that?

If you don't get an answer from @core23 before this is merged, then yes, please do.

core23:
> > supress deprecation / move service and alias
>
> => this should not block the merge request. shall i open an issue about that?

Feel free to fix this 👍

dbu:
the issue already exists => #363

lets merge this improvement?

greg0ire:
Thanks @dbu !

PR #367, opened 2017-02-27
avoid triggering our own deprecation notice

### LINKED ISSUE #363: 3.3.0 throwing strange deprecation notices

dbu:
I am targetting this branch, because the problem is with this branch.

Closes #363 

## Subject

Do not throw deprecation notices for code that is in this bundle (and will only be fixed for version 4).

## Changelog

```markdown
# Fixed
- Some unwanted deprecation notices about code we can't change until next major version have been removed
```

### DIFF
--- a/DependencyInjection/Compiler/TweakCompilerPass.php
+++ b/DependencyInjection/Compiler/TweakCompilerPass.php
@@ -40,7 +40,7 @@ class TweakCompilerPass implements CompilerPassInterface
             // Replace empty block id with service id
             if (strlen($arguments[0]) == 0) {
                 $definition->replaceArgument(0, $id);
-            } elseif ($id != $arguments[0]) {
+            } elseif ($id != $arguments[0] && 0 !== strpos('Sonata\\BlockBundle\\Block\\Service\\', $definition->getClass())) {
                 // NEXT_MAJOR: Remove deprecation notice
                 @trigger_error(
                     'Using different service id and block id is deprecated since 3.x and will be removed in 4.0.',


core23:
Please add a changelog entry for this

DependencyInjection/Compiler/TweakCompilerPass.php
```diff
@@ -40,7 +40,7 @@ public function process(ContainerBuilder $container)
             // Replace empty block id with service id
             if (strlen($arguments[0]) == 0) {
                 $definition->replaceArgument(0, $id);
-            } elseif ($id != $arguments[0]) {
+            } elseif ($id != $arguments[0] && 0 !== strpos('Sonata\\BlockBundle\\Block\\Service\\', $definition->getClass())) {
```
dbu: using the class rather than the service id. people are good at copy-pasting and might just use sonata.block.service for their own services where we want the deprecation notice

dbu:
> Please add a changelog entry for this

really? i would consider that a pedantic change. there is no behaviour
change at all, only something for continous integration tools.

greg0ire:
If people outside our group get something from it, it's not pedantic and warrants a changelog in this form : 

```
# Fixed
- Some unwanted deprecation notices about <insert whatever is right here> have been removed
```

greg0ire:
@dbu : wrote a changelog for you, I hope you can see some people might read it and think "At last! If I upgrade I'll get rid of this deprecation notices that bug me!"

dbu:
okay, fair enough.

greg0ire:
FYI deprecations is something users are really concerned about : http://stackoverflow.com/questions/40101753/deprecated-warnings-when-using-sonata-admin-bundle

dbu:
and they better be concerned. so we better not teach them that they are
irrelevant :-)

greg0ire:
My thoughts exactly!

greg0ire:
Also, [I wrote a PR to be able to avoid contributions that introduce deprecation without fixing corresponding calls](https://github.com/symfony/symfony/pull/21539), while still being able to ignore deprecations from vendors. Any review / support is very welcome!

dbu:
great one. we have configuration for the deprecation warner with the
number of deprecations triggered by vendors, which is stupid.

greg0ire:
Thanks! Maintaining a number is indeed cumbersome, and if you have several repos like in Sonata, you have to do that for every repo… not a great workflow :P

greg0ire:
Thanks @dbu

pkruithof:
Unfortunately the definition class is `%sonata.block.service.container.class%` rather than a FQCN, so this patch does not work yet :(

greg0ire:
@pkruithof you mean this parameter is not resolved at that time?

pkruithof:
That's what I mean, yes. I upgraded to this version, and the problem remained. Debugging the `$definition->getClass()` call yields this string.

greg0ire:
IIRC there is a resolve or similar method that could be used there… can you check?

pkruithof:
Yes, you can use this:

```php
$class = $container->getParameterBag()->resolveValue($definition->getClass());
if ($id != $arguments[0] && 0 !== strpos('Sonata\\BlockBundle\\Block\\Service\\', $class)) {
```

greg0ire:
Can you make a PR to contribute this?

greg0ire:
Not sure why it worked for @dbu though… unless this parameter was introduced recently? Can you check again @pkruithof ?

pkruithof:
As far as I can see there is no test for this class, so my guess would be that this change was not tested.

dbu:
yep, i don't think i really tested it.

pkruithof:
PR => https://github.com/sonata-project/SonataBlockBundle/pull/369

greg0ire:
> yep, i don't think i really tested it.

Won't blame you , I do that way too often :P

dbu:
thanks for fixing it and sorry again.

PR #377, opened 2017-04-04
fix(abstractBlockServiceTestCase): add many arguments for getMock

williamjehanne:
I am targeting this branch, because missing arguments in getMock method.

## Changelog

```markdown
### Added
- Add some arguments to `AbstractBlockServiceTestCase::getMock` method

### Changed

### Deprecated

### Removed

### Fixed

### Security
```

## To do

- [ ] Update the tests
- [ ] Add an upgrade note


jordisala1991:
This method is about making easier to move to PHPUnit 5/6. If we add those parameter (that I believe are not on createMock) we will have to change again our code using them.

If you need those, it is better to use mockBuilder directly

core23:
This is just a wrapper method for new PHPunit versions:
https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/TestCase.php#L1475

We don't need this, sorry

PR #378, opened 2017-04-14
prevent notice if $arguments is an empty array

### LINKED ISSUE #374: With Update to 3.3 blocks are broken
## Subject

the following error is thrown:

```
ContextErrorException in TweakCompilerPass.php line 44:
Notice: Undefined offset: 0
```

the loop for the service ids has an empty argument.

```
$id = string 'admin.block.stat.count' (length=22)
$arguments =array (size=0)  empty
```

## service definition

is done through annotations:

```php
/**
 * @DI\Service("admin.block.stat.count")
 * @DI\Tag("sonata.block")
 */
class StatCountBlock extends AbstractBlockService
```

wodka:
I am targeting this branch, because this error was introduced in the 3.3 version.

Closes #374

## Subject

if services are defined through annotations the arguments are modified later -> therefore the $arguments variable is an empty array at this point. which in return causes the error if [0] is accessed

## Changelog

```markdown
# Fixed
- a notice that appeared when defining services through annotations
```

### DIFF
--- a/DependencyInjection/Compiler/TweakCompilerPass.php
+++ b/DependencyInjection/Compiler/TweakCompilerPass.php
@@ -38,7 +38,7 @@ class TweakCompilerPass implements CompilerPassInterface
             $arguments = $definition->getArguments();
 
             // Replace empty block id with service id
-            if (strlen($arguments[0]) == 0) {
+            if (empty($arguments) || strlen($arguments[0]) == 0) {
                 $definition->replaceArgument(0, $id);
             } elseif ($id != $arguments[0] && 0 !== strpos(
                 $container->getParameterBag()->resolveValue($definition->getClass()),


wodka:
sry - I do not understand what you mean? The PR Subject is the same as the commit message I used. Should I change that to something else?

wodka:
Changed the commit - I modified the message and added the whole subject to the optional message

greg0ire:
> edit: I deleted the previous comment because I realized what you meant. thanks :)

I thought it was a bug, didn't see the change ^^

Great job!

jordisala1991:
It should contain a changelog, this is a patch right?

greg0ire:
I drafted one, you can improve it if you want @wodka

wodka:
changelog entry is fine for me this way, thank you!

jordisala1991:
Changed: services -> blocks

PR #383, opened 2017-06-02
Bugfix: Reordered if statements in BlockHelper

ceesvanegmond:
I am targeting this branch, because i was getting an error from [SontataTranslationBundle](https://github.com/sonata-project/SonataTranslationBundle). This bundle fires an ```sonata.block.event.sonata.admin.list.table.top``` event, which is a ```\Closure``` in the end.

## Changelog
```
Changed order of statements in the getEventListeners() method. 

### Fixed

- Issues where you pass in a \Closure class
```

## Subject

Bugfix: Fixed bug where the \Closure statement if never reached because it's already in the is_object() clause. The ```Closure``` class is an object, so therefore the third elseif statement is never reached.

PR #385, opened 2017-06-02
Fix Cannot use object of type Closure as array

alxb5:
I am targeting this branch, but it should be done on master too.

Closes #{384}

## Changelog
```markdown

### Fixed
- BlockHelper : Cannot use object of type Closure as array

```

## Subject

After upgrading to Symfony 3.3, I got this error: 
"Cannot use object of type Closure as array"

I think, we should test that $listener is a closure before testing that the type of the first key of a variable that is maybe not an array.


greg0ire:
Closing in favor of #383, but thanks!

PR #393, opened 2017-08-11
Ajax Widget

core23:
I am targeting this branch, because this is a new BC feature.

Moved from https://github.com/sonata-project/SonataDashboardBundle/pull/53

## Changelog

```markdown
### Added
- Added `AjaxBlockService` for Ajax statistics
```

## To do

- [X] Update the tests

## Subject

This PR adds a new widget block to load static data from an external source and visualize it:

bildschirmfoto 2017-08-11 um 08 37 28

### DIFF
--- a/Resources/translations/SonataBlockBundle.hu.xliff
+++ b/Resources/translations/SonataBlockBundle.hu.xliff
@@ -22,6 +22,46 @@
                 <source>sonata.block.template</source>
                 <target>Sablon</target>
             </trans-unit>
+            <trans-unit id="sonata.block.ajax">
+                <source>sonata.block.ajax</source>
+                <target>sonata.block.ajax</target>
+            </trans-unit>
+            <trans-unit id="form.label_text">
+              <source>form.label_text</source>
+              <target>form.label_text</target>
+            </trans-unit>
+            <trans-unit id="form.label_class">
+              <source>form.label_class</source>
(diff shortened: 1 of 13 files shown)


Resources/translations/SonataBlockBundle.hu.xliff
```diff
@@ -22,6 +22,46 @@
                 <source>sonata.block.template</source>
                 <target>Sablon</target>
             </trans-unit>
+            <trans-unit id="sonata.block.ajax">
+                <source>sonata.block.ajax</source>
+                <target>sonata.block.ajax</target>
```
Soullivaneuh: Not translated part should not be added. So the translator may fallback to the default language.

core23: When did we discuss that? Every someone adds a new key, we asked for this

Soullivaneuh: > Every time someone adds a new key, we asked for this

When did you decide that? :-)

No. This is not translated so it should not be added, no need to edit tons of file for nothing.

Plus, this will mask the translation not found warning and it's not good. :wink:

core23:
After thinking a little longer about this. The AJAX stuff should be handled globally in the https://github.com/sonata-project/SonataDashboardBundle.

PR #399, opened 2017-09-04
Fix build

core23:

## Subject

Try to fix the build...


core23:
It's passing :)

PR #410, opened 2017-10-23
Fix argument replacement of a non-existent one

Soullivaneuh:
I am targeting this branch, because it's a bugfix.

## Changelog

```markdown
### Fixed
- OutOfBoundsException while replacing block service default name argument
```

## Subject

#374 was not enough. If the service arguments list is empty, we can't call `replaceArgument`.

The `setArgument` method works for the both cases because it set the argument like a :pig:  without checking anything.

This is buggy since https://github.com/symfony/symfony/commit/428814b25ddd7b0bfbf56d0f4cc3879c3e9bc793 and also affect service autowiring like this:

```
services:
    _defaults:
        autowire: true
        autoconfigure: true
        public: false

    _instanceof:
        Sonata\BlockBundle\Block\Service\BlockServiceInterface:
            tags: [ sonata.block ]
```

### DIFF
--- a/DependencyInjection/Compiler/TweakCompilerPass.php
+++ b/DependencyInjection/Compiler/TweakCompilerPass.php
@@ -39,7 +39,7 @@ class TweakCompilerPass implements CompilerPassInterface
 
             // Replace empty block id with service id
             if (empty($arguments) || strlen($arguments[0]) == 0) {
-                $definition->replaceArgument(0, $id);
+                $definition->setArgument(0, $id);
             } elseif ($id != $arguments[0] && 0 !== strpos(
                 $container->getParameterBag()->resolveValue($definition->getClass()),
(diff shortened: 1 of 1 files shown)


greg0ire:
This crashes on sf 2.8.32 with

> Attempted to call an undefined method named "setArgument" of class "Symfony\Component\DependencyInjection\Definition".

greg0ire:
Workaround: set the block name instead of just `null`

Soullivaneuh:
If this cause issue with SF 2.8, then we should drop the version support as 3.4 is LTS too.

greg0ire:
> should drop

* should have dropped

If we do what you suggest, people will have no fix and won't be able to upgrade, so the correct solution is 

1. revert this
2. release
3. revert the revert , drop 2.8 and release

PR #420, opened 2017-11-02
phpstan static code analysis fixes

acrobat:
I am targeting this branch, because some code (style) fixes.

## Subject

Some code fixes discovered by phpstan's static code analysis

### DIFF
--- a/Block/Service/MenuBlockService.php
+++ b/Block/Service/MenuBlockService.php
@@ -83,7 +83,7 @@ class MenuBlockService extends AbstractAdminBlockService
             'context' => $blockContext,
         ];
 
-        if ('private' === $blockContext->getSettings('cache_policy')) {
+        if ('private' === $blockContext->getSetting('cache_policy')) {
             return $this->renderPrivateResponse($blockContext->getTemplate(), $responseSettings, $response);
         }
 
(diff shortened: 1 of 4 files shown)


Block/Service/MenuBlockService.php
```diff
@@ -83,7 +83,7 @@ public function execute(BlockContextInterface $blockContext, Response $response
             'context' => $blockContext,
         ];
 
-        if ('private' === $blockContext->getSettings('cache_policy')) {
+        if ('private' === $blockContext->getSetting('cache_policy')) {
```
greg0ire: lol nice

greg0ire: This fixes a bug, right? It's not just style...

acrobat: Yes this is indeed a bugfix!

acrobat:
I will see if I can come up with a test for the bug that was fixed in this PR. Maybe move that bugfix + tests into a separate PR, what do you think @greg0ire? Just keep the pure code style changes here and move the bugfix to another PR

greg0ire:
Either do that or do separate commits, I'm fine with both solutions.

Soullivaneuh:
Thank you @acrobat

PR #426, opened 2017-11-07
Remove trigger_error for deprecation

Soullivaneuh:
It does not work properly with the weak_vendors option and is not
mandatory.

I am targeting this branch, because it's a Travis fix.


greg0ire:
> is not mandatory.

You should probably explain why

greg0ire:
Explained it for you in the commit message

PR #440, opened 2017-11-29
Drop old Symfony code

jordisala1991:

### Subject

Removing old code that never gets executed because we dropped SF 2.3 to 2.7


src/Resources/stubs/symfony2.php
```diff
@@ -1,21 +0,0 @@
-<?php
```
jordisala1991: I don't know why this class was there

tests/Block/AbstractBlockServiceTest.php
```diff
@@ -1,23 +0,0 @@
-<?php
```
jordisala1991: Am I allowed to remove classes from tests directory?

tests/Block/AbstractBlockServiceTest.php
```diff
@@ -1,23 +0,0 @@
...
-use Sonata\BlockBundle\Test\AbstractBlockServiceTestCase;
-
-/**
- * NEXT_MAJOR: remove this class.
- *
- * @deprecated Deprecated since version 3.1. Use Sonata\BlockBundle\Test\AbstractBlockServiceTestCase instead
- */
-abstract class AbstractBlockServiceTest extends AbstractBlockServiceTestCase
-{
-}
```
greg0ire: Are you sure this is supposed to be removed?

jordisala1991: The correct file is on src directory, but I am not sure If I am allowed to remove it from here. 

One thing is for sure: https://github.com/sonata-project/SonataBlockBundle/blob/3.x/composer.json#L47

greg0ire: Oh right I didn't notice that.

src/Resources/stubs/symfony2.php
jordisala1991: Restored, this is necessary

greg0ire:
Thanks @jordisala1991 !

src/Block/Service/MenuBlockService.php
```diff
@@ -163,19 +164,12 @@ protected function getFormSettingsKeys()
...
-
-        // NEXT_MAJOR: remove SF 2.7+ BC
-        if (method_exists('Symfony\Component\Form\AbstractType', 'configureOptions')) {
-            $choices = array_flip($choices);
-
-            // choice_as_value options is not needed in SF 3.0+
-            if (method_exists('Symfony\Component\Form\FormTypeInterface', 'setDefaultOptions')) {
-                $choiceOptions['choices_as_values'] = true;
-            }
+        // choice_as_value options is not needed in SF 3.0+
```
OskarStark: should we name it with `NEXT_MAJOR` ?

jordisala1991: Probably yes, to be honest when I search for code removal I usually start with method_exists class_exists and interface_exists, then 5.3 and finally NEXT_MAJOR

PR #441, opened 2017-11-30
added class alias for phpunit testcase class

OskarStark:

## Changelog

```markdown
### Fixed
- `AbstractBlockServiceTestCase` now works with PHPUnit >= 6.0
```

## Subject

I got an error when using phpunit 6.0 and extending the `AbstractBlockServiceTestCase`.

```
Fatal error: Class 'PHPUnit_Framework_TestCase' not found in /var/www/vendor/sonata-project/block-bundle/src/Test/AbstractBlockServiceTestCase.php on line 26
```

I tried to fix it by a class alias.

### DIFF
--- a/src/Test/AbstractBlockServiceTestCase.php
+++ b/src/Test/AbstractBlockServiceTestCase.php
@@ -18,6 +18,13 @@ use Sonata\BlockBundle\Block\BlockServiceInterface;
 use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
+/*
+ * NEXT_MAJOR: remove check and extend PHPUnit\Framework\TestCase when dropping support for PHPUnit 4
+ */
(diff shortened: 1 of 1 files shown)


OskarStark:
> This looks good, but did you test it in "real life" :P ?

yes I did, but could we extend travis to get our tests running with phpunit 6?

greg0ire:
@OskarStark I'm not 100% convinced it would work with old versions of sf, especially in cases where we extend test case from it, but maybe we could try it on the other jobs?

OskarStark:
I mean only one extra/special job which uses PHP 7.0, SF 3.3 and Phpunit 6.0 ?

greg0ire:
If phpunit works 7.0 or (7.1) with phpunit 6, I say let's use it everywhere we can

greg0ire:
If it doesn't on all repos, then maybe an allowed failure job?

greg0ire:
Thanks @OskarStark !

jordisala1991:
One question here, do we need compat with PHPUnit 4 for this class?

Issue #389, opened 2017-07-15
Wrong menu alias when using sonata.block.menu


core23:
#### Sonata packages

```bash
sonata-project/admin-bundle                      3.20.1             The missing Symfony Admin Generator
sonata-project/block-bundle                      3.3.2              Symfony SonataBlockBundle
sonata-project/cache                             1.0.7              Cache library
sonata-project/cache-bundle                      2.3.1              This bundle provides caching services
sonata-project/classification-bundle             3.3.1              Symfony SonataClassificationBundle
sonata-project/core-bundle                       3.4.0              Symfony SonataCoreBundle
sonata-project/dashboard-bundle                  dev-master 1d62f37 Provides a Dashboard management through container and block services
sonata-project/datagrid-bundle                   2.2.1              Symfony SonataDatagridBundle
sonata-project/doctrine-extensions               1.0.2              Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle         3.1.6              Symfony Sonata / Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/easy-extends-bundle               2.2.0              Symfony SonataEasyExtendsBundle
sonata-project/exporter                          1.7.1              Lightweight Exporter library
sonata-project/formatter-bundle                  3.2.2              Symfony SonataFormatterBundle
sonata-project/google-authenticator              2.0.0              Library to integrate Google Authenticator into a PHP project
sonata-project/intl-bundle                       2.3.1              Symfony SonataIntlBundle
sonata-project/media-bundle                      3.5.1              Symfony SonataMediaBundle
sonata-project/news-bundle                       3.1.2              Symfony SonataNewsBundle
sonata-project/notification-bundle               3.1.0              Symfony SonataNotificationBundle
sonata-project/page-bundle                       3.5.1              This bundle provides a Site and Page management through container and block services
sonata-project/seo-bundle                        2.2.0              Symfony SonataSeoBundle
sonata-project/timeline-bundle                   3.1.1              Integrates SpyTimelineBundle into Sonata
sonata-project/user-bundle                       3.2.4              Symfony SonataUserBundle

```

#### Symfony packages

```bash
symfony-cmf/routing                              2.0.1              Extends the Symfony routing component for dynamic routes and chaining several routers
symfony-cmf/routing-bundle                       2.0.0              Symfony RoutingBundle
symfony/asset                                    v3.0.9             Symfony Asset Component
[...]

PR #341, opened 2016-10-29
Created menu manager for MenuBlockService

core23:
I am targetting this branch, because this is BC.

## Changelog

``` markdown
### Added
- Created `MenuManager` to collect all menus for the `MenuBlockService`
- Added new `sonata.block_menu` tag

### Deprecated
- Deprecated the array parameter in `MenuBlockService`in favor of the new `MenuManager`
```
## To do

- [x] Update the tests
- [x] Update the documentation
- [x] Add an upgrade note
## Subject

Instead of defining knp menu classes in the static configuration, you can now create new menu services by extening a the new `MenuBuilderInterface` and the new `sonata.block_menu` tag. A new `MenuManager` will hold all defined menu classes.

PR #442, opened 2017-11-30
Rewrote defining menu blocks

core23:
I am targeting this branch, because this is a bugfix.

Closes https://github.com/sonata-project/SonataBlockBundle/issues/389

## Changelog

```markdown
### Changed
 - Changed `MenuRegistry::add` method signature to allow string values instead of `MenuBuilderInterface`

### Deprecated
 - deprecated `sonata.block.menu` tag in favor of the existing `knp_menu.menu` tag
 - deprecated `MenuBuilderInterface` class
```

## Subject

The PR https://github.com/sonata-project/SonataBlockBundle/pull/341 introduced a menu registry which never worked properly. This PR is a bugfix, so all defined `knp_menu.menu` menus will be added to the registry and can be selected in the `MenuBlockService`.

This PR is technically a BC break, but no one could ever haved used the menu registry.

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -1,6 +1,12 @@
 UPGRADE 3.x
 ===========
 
+## Rewrote menu blocks
+
+Defining menu blocks by using a `<tag name="sonata.block.menu"/>` tag was removed, because it never worked properly.
+
+Use the existing `<tag name="knp_menu.menu" alias="app.main"/>` tag to define a menu block instead.
+
 UPGRADE FROM 3.4 to 3.5
 =======================
 
--- a/docs/reference/provided_blocks.rst
+++ b/docs/reference/provided_blocks.rst
@@ -33,14 +33,14 @@ MenuBlockService
 
 This block service displays a KNP Menu.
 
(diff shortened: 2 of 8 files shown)


greg0ire:
Please write a changelog directed at end users. Keep in mind that they probably do not care what file or class you changed, but what features you provided / what bugs you fixed. Try to answer the question "What benefit do I get from upgrading?"
More on http://keepachangelog.com

UPGRADE-3.x.md
```diff
@@ -1,6 +1,13 @@
 UPGRADE 3.x
 ===========
 
+## Rewrote menu blocks
+
+Defining menu blocks by using a `<tag name="sonata.block.menu"/>` tag was removed, because it never worked properly.
+
+Use the existing `<tag name="knp_menu.menu" alias="app.main"/>` tag to define a menu block instead.
+
```
OskarStark: one blank line is enough

core23: Fixed

core23:
IMHO we need all those information in the changelog, because we ever mention the deprecated classes. We also need the signature change, because it's a BC break.

PR #444, opened 2017-11-30
Fix check for PHPUnit 4

jordisala1991:
On my PR I changed the extended class from AbsctractBlockServiceTestCase to be TestCase. So the change made here will do nothing: https://github.com/sonata-project/SonataBlockBundle/pull/441

Instead we should check the other way around

### DIFF
--- a/src/Test/AbstractBlockServiceTestCase.php
+++ b/src/Test/AbstractBlockServiceTestCase.php
@@ -20,10 +20,10 @@ use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /*
- * NEXT_MAJOR: remove check and extend PHPUnit\Framework\TestCase when dropping support for PHPUnit 4
+ * NEXT_MAJOR: remove check when dropping support for PHPUnit 4
  */
-if (!class_exists('\PHPUnit_Framework_TestCase')) {
-    class_alias('PHPUnit\Framework\TestCase', '\PHPUnit_Framework_TestCase');
+if (!class_exists(TestCase::class)) {
+    class_alias('\PHPUnit_Framework_TestCase', TestCase::class);
 }
 
 /**


src/Test/AbstractBlockServiceTestCase.php
```diff
@@ -19,13 +19,6 @@
 use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
-/*
- * NEXT_MAJOR: remove check and extend PHPUnit\Framework\TestCase when dropping support for PHPUnit 4
- */
-if (!class_exists('\PHPUnit_Framework_TestCase')) {
-    class_alias('PHPUnit\Framework\TestCase', '\PHPUnit_Framework_TestCase');
```
greg0ire: This aliases `TestCase` to the phpunit 4 version, so that extending `TestCase` works even when that class does not exist. Why remove that?

jordisala1991: If we want to keep compat with phpunit 4, then we might instead change on what class is extending the abstracttestcase, because now it is extending directly from TestCase

greg0ire: Yeah but that provides compat with phpunit 6. So right now we might have 4,5 and 6

jordisala1991: `class_alias` create an alias for `PHPUnit_Framework_TestCase` which is not used on this file at the moment

jordisala1991:
What about doing it like this @greg0ire ?

greg0ire:
It makes more sense now

src/Test/AbstractBlockServiceTestCase.php
```diff
@@ -20,10 +20,10 @@
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /*
- * NEXT_MAJOR: remove check and extend PHPUnit\Framework\TestCase when dropping support for PHPUnit 4
+ * NEXT_MAJOR: remove check when dropping support for PHPUnit 4
  */
-if (!class_exists('\PHPUnit_Framework_TestCase')) {
-    class_alias('PHPUnit\Framework\TestCase', '\PHPUnit_Framework_TestCase');
+if (!class_exists(TestCase::class)) {
```
OskarStark: Wouldn't this fail with a used use statement which isn't available?

jordisala1991: You can try it. Last time I did, it does not matter if class exists or not

jordisala1991: https://3v4l.org/NSTLf

greg0ire: https://3v4l.org/p36qM

PR #454, opened 2017-12-14
Manual merge

greg0ire:
Don't merge, just approve

### DIFF
--- a/src/Block/Service/EmptyBlockService.php
+++ b/src/Block/Service/EmptyBlockService.php
@@ -24,16 +24,10 @@ class EmptyBlockService extends AbstractBlockService
 {
     // NEXT_MAJOR: Remove this method
 
-    /**
-     * {@inheritdoc}
-     */
     public function configureSettings(OptionsResolver $resolver): void
     {
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function buildEditForm(FormMapper $form, BlockInterface $block): void
     {
         throw new \RuntimeException('Not used, this block renders an empty result if no block document can be found');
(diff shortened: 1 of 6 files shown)


src/Block/Service/EmptyBlockService.php
```diff
@@ -18,22 +18,11 @@
 use Sonata\BlockBundle\Model\BlockInterface;
 use Sonata\CoreBundle\Validator\ErrorElement;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\OptionsResolver\OptionsResolver;
 
 class EmptyBlockService extends AbstractBlockService
 {
     // NEXT_MAJOR: Remove this method
```
jordisala1991: We did remove some method here on 3.x but not removed the comment (on 3.x is also not removed, so don't do it here either). Just a reminder

src/Block/Service/EmptyBlockService.php
```diff
@@ -18,22 +18,11 @@
...
 
-    /**
-     * {@inheritdoc}
-     */
-    public function configureSettings(OptionsResolver $resolver): void
```
jordisala1991: I am trying to find where this was removed on 3.x, can't really find. Can you show me?

greg0ire: That's probably an error on my part

src/Menu/MenuBuilderInterface.php
```diff
@@ -19,7 +19,7 @@
 /**
  * @author Christian Gripp <mail@core23.de>
  *
- * @deprecated since 3.x, to be removed with 4.0.
+ * @deprecated since 3.9, to be removed with 4.0.
```
greg0ire: I don't know why this changed, but It doesn't matter I guess, since this will be removed in the end.

jordisala1991:
This one was hard to review because one commit was not on 3.x

PR #455, opened 2017-12-20
Make services public

core23:

## Changelog

```markdown
### Changed
 - make services explicit public
 - Autoregister blocks as public services 
```

## Subject

This patch is needed for sf 4.


OskarStark:
> A testcase would be nice

I already asked @greg0ire for the same in another PR and we decided, that it does not make that much sense in this case.

OskarStark:
Thanks @core23 👍

Issue #456, opened 2017-12-25
Switch from Templating Component to Twig


covex-nn:
I believe `Templating Component` will be deprecated soon, see https://github.com/symfony/symfony/pull/21035. So, i have a plan how to switch from `Templating Component` to `Twig`. The plan has 3 stages:

- [X] `Stage 1`, stable branch `3.x`: Use `Templating Component` without `templating` service and with own classes for templating.
        See https://github.com/sonata-project/SonataBlockBundle/pull/457 and https://github.com/sonata-project/dev-kit/issues/380
- [x] `Stage 2`, `master` branch: Remove usage of `Templating Component` in code, and use `Twig` directly or own classes for templating
- [x] `Stage 3`, future `master` branch: remove proxy classes for templating.

I will describe what are stages internally.

Stage 1
---

1. Add `symfony/templating` and `symfony/asset` to composer.json

2. Change all template names to Twig namespaced syntax from `BundleNameBundle:directory:filename.html.twig` to `@BundleName/directory/filename.html.twig` everywhere

3. Create new empty interface `Sonata\BlockBundle\Templating\EngineInterface` extending `Symfony\Bundle\FrameworkBundle\Templating\EngineInterface`

4. Create new empty class `Sonata\BlockBundle\Templating\TemplateLocator` extending `Symfony\Bundle\FrameworkBundle\Templating\Loader\TemplateLocator`. Create new `sonata.templating.locator` service and deprecate this new class.

5. Create class `Sonata\BlockBundle\Templating\TemplateNameParser` extending `Symfony\Bundle\FrameworkBundle\Templating`. Create new service `sonata.templating.name_parser` and deprecate this new class.

6. Create new empty class `Sonata\BlockBundle\Templating\TwigEngine` extending `Symfony\Bundle\TwigBundle\TwigEngine` implemenging new interface `Sonata\BlockBundle\Templating\EngineInterface`. Create new service `sonata.templating` with `sonata.templating.locator` and `sonata.templating.name_parser` as arguments for its constructor.

7. Change `<argument type="service" id="templating"/>` to `<argument type="service" id="sonata.templating"/>` everywhere in `src/Resources/config/*.xml`

Stage 2
---

1. Update `Sonata\BlockBundle\Templating\EngineInterface` interface: copy body from `Symfony\Bundle\FrameworkBundle\Templating\EngineInterface`, remove `extends` and deprecate.

[...]

PR #457, opened 2017-12-25
Use Templating Component without "templating" service

covex-nn:
I am targeting this branch, because this is backward compatible feature.

See #456

## Changelog

```markdown
### Added
- Added `symfony/asset` and `symfony/templating` dependencies
- Added new service `sonata.templating` for use in place of `templating`

### Changed
- Referencing templates using Twig namespaced syntax

### Removed
- Removed tag `templating.helper` from `sonata.block.templating.helper` service
```

## Subject

This is first step to switching from `Templating Component` to using `Twig` directly. The following two steps are described in #456. This PR will allow to run `SonataBlockBundle` without activating templating extension in `framework`.

If this PR will be accepted, all other sonata bundles, that use SonataBlockBundle, could be updated too: services will have to use `sonata.templating` service instead of `templating` and templates must be referenced using Twig namespaced syntax

### DIFF
--- a/src/Templating/TemplateLocator.php
+++ b/src/Templating/TemplateLocator.php
@@ -18,5 +18,4 @@ use Symfony\Bundle\FrameworkBundle\Templating\Loader\TemplateLocator as Framewor
  */
 class TemplateLocator extends FrameworkTemplateLocator
 {
-
 }
--- a/src/Templating/TemplateLocator.php
+++ b/src/Templating/TemplateLocator.php
@@ -1,0 +1,22 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\BlockBundle\Templating;
+
+use Symfony\Bundle\FrameworkBundle\Templating\Loader\TemplateLocator as FrameworkTemplateLocator;
+
+/**
+ * @deprecated
+ */
+class TemplateLocator extends FrameworkTemplateLocator
+{
+
+}
--- a/src/Templating/TemplateLocator.php
+++ b/src/Templating/TemplateLocator.php
@@ -14,7 +14,7 @@ namespace Sonata\BlockBundle\Templating;
 use Symfony\Bundle\FrameworkBundle\Templating\Loader\TemplateLocator as FrameworkTemplateLocator;
 
 /**
- * @deprecated
+ * @deprecated since 3.9, to be removed with 4.0.
  */
 class TemplateLocator extends FrameworkTemplateLocator
 {
--- a/src/Templating/TemplateNameParser.php
+++ b/src/Templating/TemplateNameParser.php
@@ -1,0 +1,22 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\BlockBundle\Templating;
+
+use Symfony\Bundle\FrameworkBundle\Templating\TemplateNameParser as FrameworkTemplateNameParser;
+
+/**
+ * @deprecated
+ */
+class TemplateNameParser extends FrameworkTemplateNameParser
+{
+
+}
--- a/src/Templating/TemplateNameParser.php
+++ b/src/Templating/TemplateNameParser.php
@@ -18,5 +18,4 @@ use Symfony\Bundle\FrameworkBundle\Templating\TemplateNameParser as FrameworkTem
  */
 class TemplateNameParser extends FrameworkTemplateNameParser
 {
-
 }
--- a/src/Templating/TemplateNameParser.php
+++ b/src/Templating/TemplateNameParser.php
@@ -14,7 +14,7 @@ namespace Sonata\BlockBundle\Templating;
 use Symfony\Bundle\FrameworkBundle\Templating\TemplateNameParser as FrameworkTemplateNameParser;
 
 /**
- * @deprecated
+ * @deprecated since 3.9, to be removed with 4.0.
  */
 class TemplateNameParser extends FrameworkTemplateNameParser
 {
--- a/src/Test/FakeTemplating.php
+++ b/src/Test/FakeTemplating.php
@@ -11,7 +11,7 @@
 
 namespace Sonata\BlockBundle\Test;
 
-use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
+use Sonata\BlockBundle\Templating\EngineInterface;
 use Symfony\Component\HttpFoundation\Response;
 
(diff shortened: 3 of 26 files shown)


src/Templating/TemplateLocator.php
```diff
@@ -0,0 +1,21 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\BlockBundle\Templating;
+
+use Symfony\Bundle\FrameworkBundle\Templating\Loader\TemplateLocator as FrameworkTemplateLocator;
+
+/**
+ * @deprecated since 3.9, to be removed with 4.0.
```
OskarStark: why? its new

src/Templating/TemplateNameParser.php
```diff
@@ -0,0 +1,21 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\BlockBundle\Templating;
+
+use Symfony\Bundle\FrameworkBundle\Templating\TemplateNameParser as FrameworkTemplateNameParser;
+
+/**
+ * @deprecated since 3.9, to be removed with 4.0.
```
OskarStark: why? its new

src/Templating/TemplateLocator.php
covex-nn: I've created new service `sonata.templating`, to start migrating from `templating` to `twig` service without BC breaks in 3.x branch. `sonata.templating`, like `templating` service, has 3 arguments: `twig`, `sonata.templating.name_parser` and `sonata.templating.locator`. This new class is `sonata.templating.locator` service.

It is a new class, yes. It is needed in 3.x branch to be sure, that everything will be fine after deprecating of `symfony/templating` component. But this new class should not be used by other services, because it will be deleted in master branch (if PR will be merged, of course) accordingly with stage 2 of a plan, described in #456 

So, it is new and deprecated at the same time.

src/Templating/TemplateNameParser.php
covex-nn: Same as above. But this class is `sonata.templating.name_parser` service

aLkRicha:
Maybe this need to be done in CoreBundle? And then can be reused in other sonata bundles? 
@OskarStark @greg0ire @core23 @jordisala1991

greg0ire:
@aLkRicha what do you mean by reused? This migrations needs to be done on all bundles, right?

aLkRicha:
@greg0ire if we are using templating service in more then one sonata bundle, will be correct to create own templating service in base bundle (sonata core) and reuse it in block bundle, admin bundle, etc..

jordisala1991:
One question , it is BC break to directly go to the step where we use twig directly? Without removing any code just deprecating what it is needed

covex-nn:
@jordisala1991 the second argument of [`Sonata\AdminBundle\Block\AdminListBlockService`](https://github.com/sonata-project/SonataBlockBundle/blob/3.x/src/Block/Service/AbstractBlockService.php) (abstract parent class for all Blocks) is `Symfony\Bundle\FrameworkBundle\Templating\EngineInterface $templating`.

So, we cannot just change `EngineInterface` to `Twig\Environment` in 3.x branch - this is BC break. And there is nothing to deprecate here, except may be `AbstractBlockService` itself.

aLkRicha:
@covex-nn please, answer my question ☺️  https://github.com/sonata-project/SonataBlockBundle/pull/457#issuecomment-356382537

covex-nn:
@aLkRicha CoreBundle does not use `templating` service, BlockBundle does. The main purpose
of `sonata.templating` is to replace `templating` as a second argument of `AdminListBlockService` class in 3.x branch without BC breaks. If some other bundle uses `templating`, then there could be another way how to decouple from `symfony/templating` and how to use `twig` service instead of `templating`.

So, BlockBundle is a right place for a new `sonata.templating` service. May be `sonata.templating` should be in separate repository (not in SonataCoreBundle and not in SonataBlockBundle), but i am not sure about it (it is too complicated)

jordisala1991:
@covex-nn but you can remove the type hint, and check if the instance passed if from EngineInterface or TwigEnv.. and if not throw an exception.. and that is not BC break

covex-nn:
@jordisala1991 i used http://symfony.com/doc/current/contributing/code/bc.html to determine what change is allowed and what is not. Removing type hint of an argument of a public method is not allowed there. I thought it is a BC break.

Anyway, removing type hint could be done only after all template references in all bundles will be guaranteed switched from "_templating syntax_" to "_twig namespaced syntax_", i.e. only in master branch, in next major release. But i hoped to run SonataAdminBundle with Flex with in 3.x branch.

If "_removing type hint of an argument_" is not BC break, than my plan is not good enough, and there will be 2 stages, not 3.

jordisala1991:
Wow, they really consider it bc break, althought we have done it several times, and I am not sure why they consider a bc break, because it does not break anything (if done correctly) . Wdyt @greg0ire?

greg0ire:
Removing the type hint of a method is a BC break for extending classes (they may only widen it).

jordisala1991:
https://3v4l.org/s6f92

It throws a warning but it works, anyway if we consider it BC break we cannot do what I said.

greg0ire:
Warning are BC breaks... Even strict standards are. That's why classes / methods should be final

covex-nn:
I've updated issue, i added removing type hint for second argument of `Sonata\AdminBundle\Block\AdminListBlockService` in current master. Both `sonata.templating` and `twig` services will be used in blocks. Future master branch will not contain `sonata.templating` service.

greg0ire:
The build fails on php 7.0

OskarStark:
> The build fails on php 7.0

I restarted the build, as it was a race condition 👍

aLkRicha:
do we really need there `symfony/templating` dep?

covex-nn:
@aLkRicha we need it in 3.x branch, because Templating Component is still used. It will be removed in master branch, if this PR will be accepted

src/Test/FakeTemplating.php
```diff
@@ -11,7 +11,7 @@
 
 namespace Sonata\BlockBundle\Test;
 
-use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
+use Sonata\BlockBundle\Templating\EngineInterface;
```
jordisala1991: This is the only “use” that you are replackng even though there are more referencing to EngineInterface on the project. Dont we need to replace all of them?

covex-nn: We will replace all "use"s in master branch, if this PR will be accepted. Tests, that use `Sonata\BlockBundle\Test\FakeTemplating` show us, that when we use templating service, implementing `Sonata\BlockBundle\Templating\EngineInterface`, everything is fine

covex-nn: And we use `sonata.templating` instead of `templating` !

`sonata.templating` is `Sonata\BlockBundle\Templating\TwigEngine`, it implements `Sonata\BlockBundle\Templating\EngineInterface`.

jordisala1991:
Lets pull the trigger, great work @covex-nn

covex-nn:
@jordisala1991 please do not release it right now. I realized, that i should not remove tag from templating helper. I will create PR to fix that

PR #463, opened 2018-01-08
add one more dependency in composer for creating symfony-recipe

aLkRicha:
I am targeting this branch, because we are creating recipe for 3.x version.

Fix for https://github.com/symfony/recipes-contrib/pull/235

## Changelog

```markdown
### Added
- added symfony/templating dependency to composer for using symfony recipes
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
         "sonata-project/core-bundle": "^3.4",
         "symfony/form": "^2.8 || ^3.2 || ^4.0",
         "symfony/http-kernel": "^2.8 || ^3.2 || ^4.0",
+        "symfony/templating": "^2.8 || ^3.2 || ^4.0",
         "symfony/twig-bundle": "^2.8 || ^3.2 || ^4.0",
         "twig/twig": "^1.34 || ^2.0"
     },


OskarStark:
What about #457 ?

aLkRicha:
@OskarStark aha.. ok, i could close this one.

jordisala1991:
It could get merged independently from the other PR...

aLkRicha:
@OskarStark if all is ok, you can merge this one 😅

jordisala1991:
Please explain why did you added this dependency to composer on the changelog

aLkRicha:
@jordisala1991 because without this i'm getting this error:
```
Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 1
!!
!!  In FrameworkExtension.php line 217:
!!
!!    Templating support cannot be enabled as the Templating component is not ins
!!    talled.
!!
!!
!!
```

jordisala1991:
Yep, I get it, but you need to explain it on the changelog :) End users should know what does change with your PR

aLkRicha:
@jordisala1991 is it ok now?

covex-nn:
This PR will not help to create recipe. Because, you cannot configure `framework` in your recipe, you cannot add this to your config (this is against rules):

```yaml
framework:
    templating:
        engines: ["twig"]
```

So, even if you will add `symfony/templating` to composer.json, `templating` service will not be created, and you will see new error message about it

aLkRicha:
@covex-nn yes, but we can add something like that:
```
framework:                 # Remove this section
    templating:             # if you'll be using this in config/packages/framework.yaml
        engines: ['twig'] #
```

covex-nn:
@aLkRicha these lines will not be accepted with comments or without them. See discussion about session https://github.com/symfony/recipes/issues/262. Session can be enabled only with this config:

```yaml
framework:
    session: ~
```

... but they say, that we cannot do this. If session cannot be activated in recipe config, then templating cannot be activated too. Also, see **stof**'s comment about templating: https://github.com/symfony/flex/issues/247#issuecomment-353320243. He said, that "_they precisely want to discourage the usage of the Templating wrapper in Flex projects_".

aLkRicha:
@covex-nn so there is no way to create sonataadmin recipe because of session?

covex-nn:
See https://github.com/symfony/recipes/pull/333 - session will be activated by default after merge. Right now there will be error about session service

aLkRicha:
We need https://github.com/sonata-project/SonataBlockBundle/pull/457 instead of this.

PR #464, opened 2018-01-09
Make default_contexts parameter optional

covex-nn:
I am targeting this branch, because this is BC patch.

## Changelog

```markdown
### Removed
- Removed requirement for `default_contexts` config parameter
```

## Subject

`default_contexts` is not used for `AdminBundle`, context can be set manualy for each block. So, i suggest that it is not required. Also, with this PR no Flex-recipe will be needed for this bundle, because default configiration will be used.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -82,7 +82,6 @@ class Configuration implements ConfigurationInterface
                 ->end()
 
                 ->arrayNode('default_contexts')
-                    ->isRequired()
                     ->prototype('scalar')->end()
                 ->end()
 
--- a/tests/DependencyInjection/ConfigurationTest.php
+++ b/tests/DependencyInjection/ConfigurationTest.php
@@ -17,7 +17,10 @@ use Symfony\Component\Config\Definition\Processor;
 
 class ConfigurationTest extends TestCase
 {
-    public function testOptions()
+    /**
+     * @dataProvider providerContexts
+     */
+    public function testOptions($contexts)
     {
         $defaultTemplates = [
             'SonataPageBundle:Block:block_container.html.twig' => 'SonataPageBundle template',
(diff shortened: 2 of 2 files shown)


covex-nn:
@OskarStark there is only one file with reference to `default_contexts`: https://github.com/sonata-project/SonataAdminBundle/blob/3.x/docs/getting_started/installation.rst#step-3-configure-the-installed-bundles

```yaml
# app/config/config.yml
sonata_block:
    default_contexts: [cms]
    blocks:
        # enable the SonataAdminBundle block
        sonata.admin.block.admin_list:
            contexts: [admin]
        # ...
```

"cms" should be removed, and there could be comment after line with "default_contexts": `# this line can be removed after sonata-project/block-bundle >= 3.9.3`

OskarStark:
Ok, can you please add this to the admin bundle docs after the merge of this PR?

covex-nn:
Yes, sure, i will

jordisala1991:
Why this was required? or what happens if you let that field without defining?

covex-nn:
@jordisala1991 it was required since March 1, 2012. See [commit where it was added](https://github.com/sonata-project/SonataBlockBundle/commit/708559a7cc0edc3c4083c3456838ccc94a6cc2c0#diff-850942b3ba24ab03a40aaa81b6152852). Since that commit this parameter was never changed. This parameter could be defined, but it could be empty.

The only difference between current and old code is that in 2012 `sonata.block.form.type.block` service (`Sonata\BlockBundle\Form\Type\ServiceListType`) had second argument "all contexts". It's value was replaced from `Sonata\BlockBundle\DependencyInjection\SonataBlockExtension::configureForm`.

Now `Sonata\BlockBundle\DependencyInjection\SonataBlockExtension::configureForm` is just filling `contexts` for blocks, that does not have configured `contexts` at all.

If this parameter will not be required, than it's default value will be an empty array. For AdminBundle there will be no difference. For PageBundle this parameter should be equal to `[sonata_page_bundle]`; if it will be empty or if it will be equal to `[cms]` there won't be errors (i guess)

OskarStark:
> May be block.{id}.contexts should be required instead of default_contexts?

I guess not, because this settings is to allow all available contexts, otherwise you need to update all blocks if you add a new context. I think it is only there to allow a specific block in a specific context only, but I could be wrong

covex-nn:
I'm sorry. I did not know, that rebase dismisses reviews =(

covex-nn:
I've updated unit test

> what happens if you let that field without defining?

@jordisala1991 in `Sonata\BlockBundle\Tests\ConfigurationTest::testOptions` block without `contexts` parameter recieves contexts from `default_contexts` and block with defined `contexts` parameter stays with unchanged contexts.

OskarStark:
Thank you @covex-nn 👍

PR #475, opened 2018-02-04
phpstan findings

core23:

## Changelog

```markdown
### Added
- Added missing `symfony/stopwatch` dependency (for tests)

### Removed
 - Removed inherited `AbstractType::setDefaultOptions` method

### Fixed
 - Fixed wrong / missing PHPDoc
 - Fixed missing null checks
 - Fixed missing typecast for ttl
```

## Subject

Fixed some findings from phpstan. Maybe we should include this in our build.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,8 @@
         "jms/di-extra-bundle": "^1.7",
         "knplabs/knp-menu-bundle": "^2.0",
         "sonata-project/admin-bundle": "^3.22",
-        "symfony/phpunit-bridge": "^3.3 || ^4.0"
+        "symfony/phpunit-bridge": "^3.3 || ^4.0",
+        "symfony/stopwatch": "^2.8 || ^3.2 || ^4.0"
     },
     "suggest": {
         "jms/di-extra-bundle": "Annotations for Block definition",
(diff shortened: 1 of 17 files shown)


composer.json
```diff
@@ -38,7 +38,8 @@
         "jms/di-extra-bundle": "^1.7",
         "knplabs/knp-menu-bundle": "^2.0",
         "sonata-project/admin-bundle": "^3.22",
-        "symfony/phpunit-bridge": "^3.3 || ^4.0"
+        "symfony/phpunit-bridge": "^3.3 || ^4.0",
```
jordisala1991: Now that you touch this file, can you make this only 4.0?

src/Form/Type/ContainerTemplateType.php
```diff
@@ -58,14 +57,6 @@ public function getParent()
         return ChoiceType::class;
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function setDefaultOptions(OptionsResolverInterface $resolver)
```
jordisala1991: Isn't this a BC break?

core23: Yes, because symfony provides this method itself

core23: https://github.com/symfony/symfony/blob/v2.8.0/src/Symfony/Component/Form/AbstractType.php#L47

greg0ire:
Thanks @core23 !

PR #476, opened 2018-02-05
Remove dependency on symfony/templating

covex-nn:
I am targeting this branch, because this is BC break.

See #456 

## Changelog

```markdown
### Changed
- Use Twig directly in Blocks and Exception renderers
- `AbstractBlockService::getTemplating` method renamed to `getTwig`

### Removed
- Removed default null values for arguments of `AbstractBlockService` class
- Removed dependency on `symfony/templating` in `composer.json`
- Removed all temporary classes for templating
- Removed `FakeTemplating` class
```

## Subject

Switch from Templating Component to Twig, Stage 2


src/Block/Service/AbstractBlockService.php
```diff
@@ -31,21 +32,25 @@
...
     {
-        if (null === $name || null === $templating) {
+        if ($templating instanceof EngineInterface) {
             @trigger_error(
-                'The $name and $templating parameters will be required fields with the 4.0 release.',
+                'Using sonata.templating is deprecated since 4.0, use twig service directly.',
                 E_USER_DEPRECATED
             );
+        } elseif (!$templating instanceof Environment) {
+            throw new \TypeError('Argument 2 passed to '.get_class($this).'::__construct() must be instance of Twig\Environment, instance of '.get_class($templating).' given');
```
greg0ire: Please linebreak this, use `sprintf`, `__METHOD__` and `Environment::class`

src/Exception/Renderer/InlineRenderer.php
```diff
@@ -25,22 +25,22 @@
 class InlineRenderer implements RendererInterface
 {
     /**
-     * @var EngineInterface
+     * @var Environment
      */
-    protected $templating;
+    protected $twig;
```
greg0ire: Should be private

src/Templating/EngineInterface.php
```diff
@@ -13,8 +13,55 @@
...
+     * Renders a template.
+     *
+     * @param string $name       A template name or a TemplateReferenceInterface instance
+     * @param array  $parameters An array of parameters to pass to the template
+     *
+     * @throws \RuntimeException if the template cannot be rendered
+     *
+     * @return string The evaluated template as a string
+     */
+    public function render($name, array $parameters = []);
```
greg0ire: Please add return type hints

src/Block/Service/AbstractBlockService.php
```diff
@@ -63,7 +72,13 @@ public function __construct($name = null, EngineInterface $templating = null)
      */
     public function renderResponse($view, array $parameters = [], Response $response = null)
     {
-        return $this->getTemplating()->renderResponse($view, $parameters, $response);
+        if (null === $response) {
```
kunicmarko20: this can probably be changed to `$resposne = $response ?? new Response();` but I am not sure that makes it better 😄

src/Block/Service/AbstractBlockService.php
covex-nn: Fixed

src/Exception/Renderer/InlineRenderer.php
covex-nn: Fixed, `$twig` property is private in `InlineRenderer` and `InlineDebugRenderer`

src/Templating/EngineInterface.php
covex-nn: Added type hints and updated classes, implementing this interface

src/Block/Service/AbstractBlockService.php
covex-nn: Fixed. But, i searched for `?? new` string in master branches of all sonata branches. This will be for a first time, where `$obj = $obj ?? new className()` is used =)

src/Block/Service/AbstractBlockService.php
```diff
@@ -31,21 +32,29 @@
...
+     * @param string      $name
+     * @param Environment $templating
+     *
+     * @throws \TypeError when 2nd argument is not Twig\Environment
      */
-    public function __construct($name = null, EngineInterface $templating = null)
+    public function __construct($name, $templating)
     {
-        if (null === $name || null === $templating) {
+        if ($templating instanceof EngineInterface) {
```
jordisala1991: Do we need to add deprecation messages on master? can't we just move to twig direclty? wdyt @sonata-project/contributors @covex-nn ?

kunicmarko20: I think that is a better option.

src/Block/Service/MenuBlockService.php
```diff
@@ -57,11 +57,11 @@ class MenuBlockService extends AbstractAdminBlockService
 
     /**
      * @param string                     $name
-     * @param EngineInterface            $templating
+     * @param Environment                $templating
      * @param MenuProviderInterface      $menuProvider
      * @param MenuRegistryInterface|null $menuRegistry
      */
-    public function __construct($name, EngineInterface $templating, MenuProviderInterface $menuProvider, $menuRegistry = null)
+    public function __construct($name, $templating, MenuProviderInterface $menuProvider, $menuRegistry = null)
```
kunicmarko20: Can you replace parameters with typehint and remove the comment?

src/Exception/Renderer/InlineDebugRenderer.php
```diff
@@ -46,14 +41,19 @@ class InlineDebugRenderer implements RendererInterface
...
+    private $twig;
+
+    /**
+     * @param Environment $twig       Templating engine
+     * @param string      $template   Template to render
+     * @param bool        $debug      Whether the debug is enabled or not
+     * @param bool        $forceStyle Whether to force style within the template or not
      */
-    public function __construct(EngineInterface $templating, $template, $debug, $forceStyle = true)
+    public function __construct(Environment $twig, $template, $debug, $forceStyle = true)
```
kunicmarko20: Can you replace parameters with typehint and remove the comment?

src/Exception/Renderer/InlineRenderer.php
```diff
@@ -25,22 +25,22 @@
...
+    private $twig;
 
     /**
-     * @param EngineInterface $templating Templating engine
-     * @param string          $template   Template to render
+     * @param Environment $twig     Twig
+     * @param string      $template Template to render
      */
-    public function __construct(EngineInterface $templating, $template)
+    public function __construct(Environment $twig, $template)
```
kunicmarko20: Can you replace parameters with typehint and remove the comment?

src/Block/Service/AbstractBlockService.php
covex-nn: At the begining i thought, that if i will remove type hint from `__construct` of AbstractBlockService, it will be easier to updgrade all other bundles, because they will be compatible with 3.x and with dev-master of BlockBundle.

I searched in `src/Block` directory inside each bundle and found, that there are 12 bundles, containing `Blocks`:
1. 7 bundles contain at least ont `Block` with extended `__construct` (with more than 2 parameters): AdminBundle, DoctrineORMAdminBundle, MediaBundle, NewsBundle, PageBundle, SeoBundle, TimelineBundle.
2. 4 bundles have `Blocks` without `__construct`: CommentBundle: DashboardBundle, FormatterBundle, TranslationBundle
3. DoctrinePhpcrAdminBundle contain `Block`, that extends deprecated BaseBlockService class

So, first 7 bundles will not be compatible with new AbstractBlockService::__construct =(( May be it is too much and a should change `EngineInterface $templating` to `Environment $templating`?

src/Block/Service/MenuBlockService.php
covex-nn: @kunicmarko20 i am sorry, i did not understans what you mean =( you say, that i have to replace `$templating` with `Environment $templating`? But exception renderers contain this type hint and your comment is the same there.

kunicmarko20: @covex-nn sorry, my bad, here you can skip it but on other places you can replace comment with typehint in constructor, right?

src/Exception/Renderer/InlineDebugRenderer.php
covex-nn: But type hint is already exists here. Removing type hint for AbstractBlockService was made for compability with other bundles (for easier upgrade). Should i remove type hint from `Environment $twig` or comment `@param Environment $twig       Templating engine` ?

kunicmarko20: My comment was badly explained, I updated it. So we can typehint all parameters and remove the phpdoc comment.

covex-nn: Fixed. It is `public function __construct(Environment $twig, string $template, bool $debug, bool $forceStyle = true)` now without phpdoc comment

src/Exception/Renderer/InlineRenderer.php
covex-nn: Fixed. It is `public function __construct(Environment $twig, string $template)` now without phpdoc comment

src/Exception/Renderer/InlineDebugRenderer.php
kunicmarko20: Great, thank you very much!

src/Test/FakeTemplating.php
```diff
@@ -20,6 +20,8 @@
  * Mocking class for template usage.
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * @deprecated since 4.0, to be removed with 5.0.
  */
 class FakeTemplating implements EngineInterface
```
covex-nn: This class is not used any more. Can i remove it instead of deprecating?

src/Block/Service/AbstractBlockService.php
kunicmarko20: Well, 4.x does not have to be BC with 3.x? /cc @sonata-project/contributors

src/Test/FakeTemplating.php
jordisala1991: Yes

src/Block/Service/AbstractBlockService.php
```diff
@@ -63,7 +52,11 @@ public function __construct($name = null, EngineInterface $templating = null)
      */
     public function renderResponse($view, array $parameters = [], Response $response = null)
     {
-        return $this->getTemplating()->renderResponse($view, $parameters, $response);
+        $response = $response ?? new Response();
+
+        $response->setContent($this->getTemplating()->render($view, $parameters));
```
jordisala1991: setContent returns the response, so you can return directly if you want :P

src/Templating/EngineInterface.php
```diff
@@ -13,8 +13,55 @@
 
 namespace Sonata\BlockBundle\Templating;
 
-use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface as TemplatingEngineInterface;
+use Symfony\Component\HttpFoundation\Response;
 
-interface EngineInterface extends TemplatingEngineInterface
+/**
+ * @deprecated since 4.0, to be removed with 5.0.
```
jordisala1991: If not used, you can remove it also

covex-nn: It is not used =)
If so, i will remove templating helper too! All methods will be copied to twig extension

src/Block/Service/AbstractBlockService.php
covex-nn: I do not like it =) with these two lines i can set breakpoint on a second line with `return $response` after setContent; with one line it won't be convenient

src/Block/Service/AbstractBlockService.php
greg0ire: If I understand correctly, the issue is that some bundle need to be compatible with 2 versions of some other bundles, right?

src/Block/Service/AbstractBlockService.php
```diff
@@ -31,23 +31,12 @@
...
     {
-        if (null === $name || null === $templating) {
-            @trigger_error(
-                'The $name and $templating parameters will be required fields with the 4.0 release.',
-                E_USER_DEPRECATED
-            );
-        }
-
         $this->name = $name;
         $this->templating = $templating;
```
jordisala1991: Can this be renamed to twig? And maybe the getTemplate method too? it will be a lof of changes maybe so I am not sure, but it seems wrong

covex-nn: @jordisala1991 done
Build failed, because 41 is not 42 again =) Please, restart travis job https://travis-ci.org/sonata-project/SonataBlockBundle/jobs/341291098

composer.json
```diff
@@ -34,7 +34,6 @@
         "symfony/http-foundation": "^2.8 || ^3.2 || ^4.0",
         "symfony/http-kernel": "^2.8 || ^3.2 || ^4.0",
         "symfony/options-resolver": "^2.8 || ^3.2 || ^4.0",
-        "symfony/templating": "^2.8 || ^3.2 || ^4.0",
```
jordisala1991: Can you check if we can get rid of framework-bundle also? AFAIK it was used for the EngineInterface, not sure if we need it for anything else

tests/Exception/Renderer/InlineDebugRendererTest.php
```diff
@@ -33,7 +33,7 @@ public function testRenderWithoutDebug(): void
         $debug = false;
         $exception = $this->createMock('\Exception');
         $block = $this->createMock('Sonata\BlockBundle\Model\BlockInterface');
-        $templating = $this->createMock('Symfony\Bundle\FrameworkBundle\Templating\EngineInterface');
+        $templating = $this->createMock('Twig\Environment');
```
jordisala1991: Can you use ::class notation? Only on the lines you changed. No need to fix the whole file

composer.json
covex-nn: `symfony/framework-bundle` is used only in `Sonata\BlockBundle\Command\BaseCommand`:

```php
namespace Sonata\BlockBundle\Command;

use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;

abstract class BaseCommand extends ContainerAwareCommand { /* ... */ }
```

To remove dependency `extends ContainerAwareCommand` could be replaced with `implements \Symfony\Component\DependencyInjection\ContainerAwareInterface` here

tests/Exception/Renderer/InlineDebugRendererTest.php
covex-nn: Done. Also i renamed all `$templating` to `$twig` in test classes

covex-nn:
@jordisala1991 i've updated upgrading note and some docs. But this is the most difficult part of PR =)

UPGRADE-4.0.md
```diff
@@ -12,3 +12,9 @@ See also the [diff code](https://github.com/sonata-project/SonataAdminBundle/com
 ## Block id
 
 If you have created a custom `AbstractBlockService` you must now implement the new constructor, because all blocks use the service id as the block id now. Because of that, the translation keys for the block ids have changed from `sonata.block.$NAME` to `sonata.block.service.$NAME`.
+
+## Block constructor
+
+Blocks are using `twig` service to render templates, so if you already overridden a constructor of a custom `AbstractBlockService`, you must update it.
```
greg0ire: "overrode"

covex-nn: Fixed

kunicmarko20:
ping @sonata-project/contributors can you review this and maybe get it merged.

OskarStark:
IMO 3 approvals are enough 👍

OskarStark:
Thank you very much @covex-nn 👍

PR #481, opened 2018-02-19
Make services public

kunicmarko20:
I am targeting this branch, because this is a BC fix.

## Changelog

```markdown
### Fixed
- sonata.block.manager is public
```

### DIFF
--- a/src/Resources/config/core.xml
+++ b/src/Resources/config/core.xml
@@ -13,7 +13,7 @@
             <argument type="service" id="sonata.templating.name_parser"/>
             <argument type="service" id="sonata.templating.locator"/>
         </service>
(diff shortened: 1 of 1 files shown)


jordisala1991:
For anyone wondering, its usage (there might be more out there): https://github.com/sonata-project/SonataBlockBundle/blob/7522bde1887baf8ea14c6e31f9d8cfb3b95c2f50/src/Command/BaseCommand.php#L24

kunicmarko20:
I encountered the error in page, https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Controller/PageAdminController.php#L245 :smile:

PR #506, opened 2018-10-04
Moved classes from CoreBundle

core23:

## Changelog

```markdown
### Added
- Added `Meta\Metadata` class (import from CoreBundle)
- Added `Meta\MetadataInterface` class (import from CoreBundle)
```

## Subject

An other step to get rid of the CoreBundle (https://github.com/sonata-project/dev-kit/issues/348)


src/Meta/Metadata.php
```diff
@@ -0,0 +1,89 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\BlockBundle\Meta;
+
+/**
+ * @author Hugo Briand <briand@ekino.com>
+ */
+class Metadata implements MetadataInterface
```
kunicmarko20: should this extend the core one? What if someone is depending on the actual concrete class.

core23: General speaking I'm with you, but we don't replace any service and the user can decide which implementation (CoreBundle oder BlockBundle) of `MetadataInterface` we wants.

core23:
@kunicmarko20 Can we merge this?

PR #508, opened 2018-10-12
Autoregister blocks

### LINKED ISSUE #471: Autoregister blocks
Today you have to register every newly created block under `sonata_block.blocks` (https://github.com/sonata-project/SonataBlockBundle/blob/3.x/src/DependencyInjection/Configuration.php#L127).

This is really annoying. So, what do you think about an autoregistration. Is there any case where you don't want to register a block that is tagged with `sonata.block`?

The same logic could be applied to breadcrumbs tagged with `sonata.breadcrumb`.

core23:
I am targeting this branch, because this feature is BC

Closes #471

## Changelog

```markdown
### Added
- All blocks containing `sonata.block` will be auto registered
```

## To do
- [x] Update the tests
- [x] Update the documentation

## Subject

You don't have to add all blocks to the config anymore, but you can to override the auto config.

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -4,6 +4,12 @@ UPGRADE 3.x
 UPGRADE FROM 3.12 to 3.x
 =======================
 
+## Autoregister blocks
+
+All blocks tagged with `sonata.block` will now be autoregistered to the `sonata.block.loader.service`.
+
+You can still override the block configuration with `sonata_block.blocks`, e.g. to change the contexts.
+
 ## Deprecate block assets 
 
 Deprecated `BlockServiceInterface::getJavascripts` and `BlockServiceInterface::getStylesheets`. 
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -6,7 +6,7 @@ UPGRADE FROM 3.12 to 3.x
 
 ## Autoregister blocks
 
-All blocks tagged with `sonata.block` will now be autoregistered to the `sonata.block.loader.service`.
+All blocks tagged with `sonata.block` will now be autoregistered into the `sonata.block.loader.service`.
 
 You can still override the block configuration with `sonata_block.blocks`, e.g. to change the contexts.
 
--- a/docs/reference/installation.rst
+++ b/docs/reference/installation.rst
@@ -68,19 +68,8 @@ To use the ``BlockBundle``, add the following lines to your application configur
         sonata_block:
             default_contexts: [sonata_page_bundle]
             blocks:
-                sonata.admin.block.admin_list:
-                    contexts:   [admin]
-
-                #sonata.admin_doctrine_orm.block.audit:
-                #    contexts:   [admin]
-
                 sonata.block.service.text:
                 sonata.block.service.rss:
-
-                # Some specific block from the SonataMediaBundle
-                #sonata.media.block.media:
-                #sonata.media.block.gallery:
-                #sonata.media.block.feature_media:
 
                 # Some block with different templates
                 #acme.demo.block.demo:
--- a/docs/reference/installation.rst
+++ b/docs/reference/installation.rst
@@ -68,9 +68,6 @@ To use the ``BlockBundle``, add the following lines to your application configur
         sonata_block:
             default_contexts: [sonata_page_bundle]
             blocks:
-                sonata.block.service.text:
-                sonata.block.service.rss:
-
                 # Some block with different templates
                 #acme.demo.block.demo:
                 #    templates:
--- a/src/DependencyInjection/Compiler/TweakCompilerPass.php
+++ b/src/DependencyInjection/Compiler/TweakCompilerPass.php
@@ -142,9 +142,9 @@ class TweakCompilerPass implements CompilerPassInterface
      */
     private function getContextFromTags(array $tags)
     {
-        return array_filter(array_map(function ($attribute) {
+        return array_filter(array_map(function (array $attribute) {
             if (array_key_exists('context', $attribute) && $attribute['context']) {
-                return $attribute['context'];
+                return (string) $attribute['context'];
             }
 
             return null;
--- a/src/DependencyInjection/Compiler/TweakCompilerPass.php
+++ b/src/DependencyInjection/Compiler/TweakCompilerPass.php
@@ -143,8 +143,8 @@ class TweakCompilerPass implements CompilerPassInterface
     private function getContextFromTags(array $tags)
     {
         return array_filter(array_map(function (array $attribute) {
-            if (array_key_exists('context', $attribute) && $attribute['context']) {
-                return (string) $attribute['context'];
+            if (array_key_exists('context', $attribute) && \is_string($attribute['context'])) {
+                return $attribute['context'];
             }
 
             return null;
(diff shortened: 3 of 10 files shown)


src/Util/StringUtil.php
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\BlockBundle\Util;
+
+class StringUtil
```
greg0ire: How about `Sonata\BlockBundle\BlockName\ConvertFromFqcn::__invoke` instead, with that class final? That kind of "util" class often ends up growing to the infinite, and being cluttered with unrelated methods whose which have only the `string` => `string` signature in common.

src/DependencyInjection/Compiler/TweakCompilerPass.php
```diff
@@ -99,4 +118,18 @@ public function applyContext(ContainerBuilder $container)
             }
         }
     }
+
+    /**
+     * @return array
```
greg0ire: Please make this more precise

tests/Util/StringUtilTest.php
```diff
@@ -0,0 +1,44 @@
...
+use Sonata\BlockBundle\Util\StringUtil;
+
+class StringUtilTest extends TestCase
+{
+    /**
+     * @dataProvider fqcnToBlockNameProvider
+     */
+    public function testFqcnToBlockName($fqcn, $expectedBlockName)
+    {
+        $blockName = StringUtil::fqcnToBlockName($fqcn);
```
greg0ire: No need for that variable IMO

src/Util/StringUtil.php
```diff
@@ -0,0 +1,37 @@
...
+    private function __construct()
+    {
+    }
+
+    /**
+     * Converts a fully-qualified class name to a block name.
+     *
+     * @param string $fqcn The fully-qualified class name
+     *
+     * @return string|null The block name or null if not a valid FQCN
```
greg0ire: Why not throw on invalid FQCN?

UPGRADE-3.x.md
```diff
@@ -1,6 +1,15 @@
 UPGRADE 3.x
 ===========
 
+UPGRADE FROM 3.12 to 3.x
+=======================
+
+## Autoregister blocks
+
+All blocks tagged with `sonata.block` will now be autoregistered to the `sonata.block.loader.service`.
```
greg0ire: "to" => "into"

src/Naming/ConvertFromFqcn.php
```diff
@@ -0,0 +1,37 @@
...
+namespace Sonata\BlockBundle\Naming;
+
+use Sonata\BlockBundle\Naming\Exception\NamingException;
+
+/**
+ * Converts a fully-qualified class name to a block name.
+ *
+ * @author Christian Gripp <mail@core23.de>
+ */
+class ConvertFromFqcn
```
greg0ire: Please make it `final`

src/Naming/ConvertFromFqcn.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    /**
+     * @param string $fqcn The fully-qualified class name
+     *
+     * @return string The block name or null if not a valid FQCN
```
greg0ire: This comment is no longer accurate

docs/reference/installation.rst
```diff
@@ -68,20 +68,6 @@ To use the ``BlockBundle``, add the following lines to your application configur
         sonata_block:
             default_contexts: [sonata_page_bundle]
             blocks:
-                sonata.admin.block.admin_list:
```
kunicmarko20: https://github.com/symfony/recipes-contrib/blob/master/sonata-project/admin-bundle/3.31/config/packages/sonata_admin.yaml

should probably be changed here also

src/Naming/ConvertFromFqcn.php
```diff
@@ -0,0 +1,39 @@
...
+     * @return string The block name
+     */
+    public function __invoke($fqcn)
+    {
+        // Non-greedy ("+?") to match "service" suffix, if present
+        if (preg_match('~([^\\\\]+?)(service)?$~i', $fqcn, $matches)) {
+            return strtolower(preg_replace(['/([A-Z]+)([A-Z][a-z])/', '/([a-z\d])([A-Z])/'], ['\\1_\\2', '\\1_\\2'], $matches[1]));
+        }
+
+        throw new NamingException(sprintf('The name "%s" is not a valid FCQN', $fqcn));
```
OskarStark: You didn’t catch it somewhere, so why don’t use InvalidArgumentException?

If not InvalidArgumentException, maybe InvalidBlockIdException ??

tests/Naming/ConvertFromFqcnTest.php
```diff
@@ -0,0 +1,45 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\BlockBundle\Tests\Naming;
+
+use PHPUnit\Framework\TestCase;
+use Sonata\BlockBundle\Naming\ConvertFromFqcn;
+
+class ConvertFromFqcnTest extends TestCase
```
OskarStark: Please add a testcase for the exception case

src/Naming/ConvertFromFqcn.php
core23: The service has nothing todo with block ids. It's just about naming blocks.

src/DependencyInjection/Compiler/TweakCompilerPass.php
```diff
@@ -116,4 +136,18 @@ private function replaceBlockName(ContainerBuilder $container, Definition $defin
             );
         }
     }
+
+    /**
+     * @return string[]
+     */
+    private function getContextFromTags(array $tags)
+    {
+        return array_filter(array_map(function ($attribute) {
```
greg0ire: Missing type hints

src/DependencyInjection/Compiler/TweakCompilerPass.php
```diff
@@ -116,4 +136,18 @@ private function replaceBlockName(ContainerBuilder $container, Definition $defin
             );
         }
     }
+
+    /**
+     * @return string[]
```
greg0ire: Are you sure? When reading the code, I'm not under the impression that `$attribute` is a string.

core23: Yes, it's a string, if you declate a `<tag ... context="foo">` inside your `services.xml`

core23: No, the method should return just a string array

greg0ire: Ooh right my bad, I thought this was an `@param` phpdoc! Can you please add it?

greg0ire:
```suggestion
     * @param string[][]
     *
     * @return string[]
```

greg0ire:
Thanks @core23 !

PR #509, opened 2018-10-18
Fix tests following convention's project naming

### LINKED ISSUE #425: AbstractBlockServiceTestCase should be renamed to BlockServiceTestCase
Test case are should not prefixed by the `Abstract` key, like this: https://github.com/symfony/symfony/blob/ba3b17719272e614732a3d1323b556cf32da8dfb/src/Symfony/Component/Form/Test/TypeTestCase.php#L18

As it's already a moved, we may deprecate this one after 4.0 release.

Th3Mouk:
I am targeting this branch, because it don't break anything.

Closes #425

## Changelog

```markdown

### Changed
- 1 class name in tests

```

### DIFF
--- a/src/Test/BlockServiceTestCase.php
+++ b/src/Test/BlockServiceTestCase.php
@@ -24,7 +24,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  *
  * @author Sullivan Senechal <soullivaneuh@gmail.com>
  */
-abstract class AbstractBlockServiceTestCase extends TestCase
+abstract class BlockServiceTestCase extends TestCase
 {
     /**
      * @var \PHPUnit_Framework_MockObject_MockObject|ContainerInterface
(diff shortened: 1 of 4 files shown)


src/Test/BlockServiceTestCase.php
```diff
@@ -24,7 +24,7 @@
  *
  * @author Sullivan Senechal <soullivaneuh@gmail.com>
  */
-abstract class AbstractBlockServiceTestCase extends TestCase
+abstract class BlockServiceTestCase extends TestCase
```
kunicmarko20: This is a BC Break, this file is in `src`

core23: You should introduce a wrapper class and deprecate `AbstractBlockServiceTestCase`

OskarStark:
Any news here?

OskarStark:
Lets close it then.

Feel free to reopen any time 👌🏻

PR #512, opened 2018-10-21
Allow autowiring blocks

core23:

## Changelog

```markdown
### Fixed
- Allow autowiring blocks
```

## Subject

You couldn't use autowiring, because the Compiler Pass tries to fill in the service name every time.

### DIFF
--- a/src/DependencyInjection/Compiler/TweakCompilerPass.php
+++ b/src/DependencyInjection/Compiler/TweakCompilerPass.php
@@ -93,13 +93,19 @@ class TweakCompilerPass implements CompilerPassInterface
 
         // Replace empty block id with service id
         if (empty($arguments) || 0 == \strlen($arguments[0])) {
-            // NEXT_MAJOR: Remove the condition when Symfony 2.8 support will be dropped.
+            // NEXT_MAJOR: Remove the if block when Symfony 2.8 support will be dropped.
             if (method_exists($definition, 'setArgument')) {
                 $definition->setArgument(0, $id);
-            } else {
-                $definition->replaceArgument(0, $id);
+
+                return;
             }
-        } elseif ($id != $arguments[0] && 0 !== strpos(
+
+            $definition->replaceArgument(0, $id);
+
+            return;
+        }
+
+        if ($id != $arguments[0] && 0 !== strpos(
             $container->getParameterBag()->resolveValue($definition->getClass()),
             'Sonata\\BlockBundle\\Block\\Service\\'
         )) {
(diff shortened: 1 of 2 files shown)


OskarStark:
Could you add a Testcase?

tests/DependencyInjection/Compiler/TweakCompilerPassTest.php
```diff
@@ -0,0 +1,115 @@
...
+            $this->markTestSkipped(
+              'Autowiring does not exist in < symfony 3.'
+            );
+        }
+
+        $blockDefinition = new Definition(null, ['acme.block.service']);
+        $blockDefinition->addTag('sonata.block');
+        $blockDefinition->setAutoconfigured(true);
+
+        $managerDefintion = $this->createMock(Definition::class);
```
greg0ire: "Definition", here and below

core23:
> Feels like this could be a BC break, does it fail if you try to use autowire?

Yes, it fails if you have a custom service which doesn't have a string as first constructor argument

src/DependencyInjection/Compiler/TweakCompilerPass.php
```diff
@@ -99,4 +83,31 @@ public function applyContext(ContainerBuilder $container)
...
+            $container->getParameterBag()->resolveValue($definition->getClass()),
+            'Sonata\\BlockBundle\\Block\\Service\\'
+        )) {
+            // NEXT_MAJOR: Remove deprecation notice
+            @trigger_error(
+                sprintf('Using service id %s different from block id %s is deprecated since 3.3 and will be removed in 4.0.', $id, $arguments[0]),
+                E_USER_DEPRECATED
+            );
+        }
+    }
```
greg0ire: How about this?
```php
<?php
$arguments = $definition->getArguments();
// Replace empty block id with service id
if (empty($arguments) || 0 == \strlen($arguments[0])) {
    // NEXT_MAJOR: Remove the if block when Symfony 2.8 support will be dropped.
    if (method_exists($definition, 'setArgument')) {
        $definition->setArgument(0, $id);
        return;
    }
    $definition->replaceArgument(0, $id);
    return;
}
if ($id != $arguments[0] && 0 !== strpos(
    $container->getParameterBag()->resolveValue($definition->getClass()),
    'Sonata\\BlockBundle\\Block\\Service\\'
)) {
    // NEXT_MAJOR: Remove deprecation notice
    @trigger_error(
        sprintf('Using service id %s different from block id %s is deprecated since 3.3 and will be removed in 4.0.', $id, $arguments[0]),
        E_USER_DEPRECATED
    );
}
```

greg0ire:
Thanks @core23 !

PR #513, opened 2018-10-30
Decouple bundle from AdminBundle

### LINKED ISSUE #493: Remove coupling to admin bundle
## Feature Request

There is a circular dependency between this and the admin bundle
https://github.com/sonata-project/SonataBlockBundle/blob/3.x/composer.json#L47
https://github.com/sonata-project/SonataAdminBundle/blob/3.x/composer.json#L29

We should remove this coupling by either:
- Move the FormMapper to the this bundle
- Use the ports and adapter pattern to define an own FormMapper interface in this bundle and an adapter in the admin bundle.

core23:

Closes #493

## Changelog

```markdown
### Added
- Added `EditableBlockService` and `FormMapper` interfaces
```

## Subject

This is the first step to decouple this bundle from the AdminBundle by inversing the dependencies. After merging this, we need to create an adapter in the AdminBundle.

Depends on https://github.com/sonata-project/form-extensions/pull/7/files

### DIFF
--- a/src/Block/Service/EditableBlockService.php
+++ b/src/Block/Service/EditableBlockService.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\BlockBundle\Block\Service;
+
+use Sonata\BlockBundle\Form\Mapper\FormMapper;
+use Sonata\BlockBundle\Meta\MetadataInterface;
+use Sonata\BlockBundle\Model\BlockInterface;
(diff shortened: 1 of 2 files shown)


src/Block/Service/EditableBlockService.php
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * @param FormMapper     $form
+     * @param BlockInterface $block
+     */
+    public function configureCreateForm(FormMapper $form, BlockInterface $block);
+
+    /**
+     * @param ErrorElement   $errorElement
+     * @param BlockInterface $block
```
kunicmarko20: this is not needed same for comments before

kunicmarko20:
I don't see any kind of implementation here?

core23:
> I don't see any kind of implementation here?

That's right. Either you have an own implementation, if you need the editable stuff or you can use the admin bundle which will support an adapter for this.

greg0ire:
Thanks @core23!

PR #514, opened 2018-11-01
Applied fixes from FlintCI

Soullivaneuh:


OskarStark:
Hmm why wasn’t this fixed before on the PR? @core23

core23:
Cause we don't require this in the stable branch

PR #553, opened 2019-02-04
Improve performance and entropy thanks to $more_entropy in uniqid()

jacquesbh:
## Use `$more_entropy = true` when calling `uniqid()`: faster and better

I was running a project which uses this bundle. I found that it uses `uniqid()` a few times. By changing the second parameter to `true` the entropy is safer and the execution is faster.

So, I did it everywhere I found `uniqid()`. It doesn't change the result of the method, it's just faster and the entropy is better. So, why not?

It's a simple patch, should be on `3.x`.

## Changelog

```markdown
### Changed
- Improve performance and entropy when calling `uniqid` from [@jacquesbh](https://github.com/jacquesbh).
```

### DIFF
--- a/docs/reference/events.rst
+++ b/docs/reference/events.rst
@@ -55,7 +55,7 @@ The `event listener` must push one or some ``BlockInterface`` instances into ``B
         public function onBlock(BlockEvent $event)
         {
             $block = new Block();
-            $block->setId(uniqid()); // set a fake id
+            $block->setId(uniqid('', true)); // set a fake id
             $block->setSettings($event->getSettings());
             $block->setType('sonata.comment.block.discus');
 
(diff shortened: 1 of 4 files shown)


OskarStark:
Not sure it’s worth, we are not doing 100000 calls against it. 
Can you show some blackfire profiles in a real sonata project?

lyrixx:
Could you elaborate why you think this should not be merged ?
I could not agree more with Jacques !

kunicmarko20:
I don't see any issue with merging this, but as @OskarStark said, there is no way this is even noticeable.

greg0ire:
Regarding the author in the changelog, I never saw that in our changelogs, but apparently it's something that can happen in [the template we are following](https://keepachangelog.com/en/1.0.0/) and would be a nice way to say thanks

jacquesbh:
@greg0ire I can remove it if you prefer.

OskarStark:
Totally fine with merging it. Could we automate it for the future by providing/using a PHP-CS-Fixer rule? 🤔

Thank you for the contribution 🙏

greg0ire:
Thanks @jacquesbh !

PR #562, opened 2019-03-19
Extend symfony DataCollector, do not serialize created_at as \DateTime

### LINKED ISSUE #556: Errors with deserialization for the profiling of blocks in debug mode with PHP 7.3
#### Sonata packages

```
$ composer show --latest 'sonata-project/*'
sonata-project/admin-bundle              3.45.2 3.45.2 The missing Symfony Admin Generator
sonata-project/block-bundle              3.14.0 3.14.0 Symfony SonataBlockBundle
sonata-project/cache                     2.0.1  2.0.1  Cache library
sonata-project/classification-bundle     3.8.0  3.8.0  Symfony SonataClassificationBundle
sonata-project/core-bundle               3.16.0 3.16.0 Symfony SonataCoreBundle
sonata-project/datagrid-bundle           2.4.0  2.4.0  Symfony SonataDatagridBundle
sonata-project/doctrine-extensions       1.1.5  1.1.5  Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.8.2  3.8.2  Symfony Sonata / Integrate Doctrine ORM into the Sona...
sonata-project/easy-extends-bundle       2.5.0  2.5.0  Symfony SonataEasyExtendsBundle
sonata-project/exporter                  2.0.1  2.0.1  Lightweight Exporter library
sonata-project/media-bundle              3.18.1 3.18.1 Symfony SonataMediaBundle
sonata-project/notification-bundle       3.5.1  3.5.1  Symfony SonataNotificationBundle

```

#### Symfony packages

```
$ composer show --latest 'symfony/*'
symfony/monolog-bundle     v3.3.1             v3.3.1             Symfony MonologBundle
symfony/phpunit-bridge     v3.4.22            v4.2.3             Symfony PHPUnit Bridge
symfony/polyfill-apcu      v1.10.0            v1.10.0            Symfony polyfill backporting apcu_* functio...
symfony/polyfill-ctype     v1.10.0            v1.10.0            Symfony polyfill for ctype functions
symfony/polyfill-intl-icu  v1.10.0            v1.10.0            Symfony polyfill for intl's ICU-related dat...
symfony/polyfill-mbstring  v1.10.0            v1.10.0            Symfony polyfill for the Mbstring extension
symfony/polyfill-php56     v1.10.0            v1.10.0            Symfony polyfill backporting some PHP 5.6+ ...
symfony/polyfill-php70     v1.10.0            v1.10.0            Symfony polyfill backporting some PHP 7.0+ ...
symfony/polyfill-util      v1.10.0            v1.10.0            Symfony utilities for portability of PHP codes
symfony/security-acl       v3.0.1             v3.0.1             Symfony Security Component - ACL (Access Co...
symfony/symfony            v3.4.22            v4.2.3             The Symfony PHP framework
```

#### PHP version

```
$ php -v
PHP 7.3.2-3+0~20190208150759.31+jessie~1.gbp0912bd (cli) (built: Feb  8 2019 23:17:46) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.3.2, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.3.2-3+0~20190208150759.31+jessie~1.gbp0912bd, Copyright (c) 1999-2018, by Zend Technologies
    with Xdebug v2.7.0rc1, Copyright (c) 2002-2019, by Derick Rethans
```

## Subject

An error occured when the profiler is retrieving data from blocks.

## Steps to reproduce

Execute any page with a SonataBlock on development environment.

## Expected results

Get information about Sonata blocks from the page in the profiler.

## Actual results

```
[2019-02-16 17:08:17] [82414226] request.CRITICAL: Uncaught PHP Exception Symfony\Component\Debug\Exception\ContextErrorException: "Notice: unserialize(): Error at offset 7078 of 12086 bytes" at /var/www/cms/vendor/sonata-project/block-bundle/src/Profiler/DataCollector/BlockDataCollector.php line 167 {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\ContextErrorException(code: 0): Notice: unserialize(): Error at offset 7078 of 12086 bytes at /var/www/cms/vendor/sonata-project/block-bundle/src/Profiler/DataCollector/BlockDataCollector.php:167)"}\
```
There is a bug in serialization with PHP 7.3 : 
https://bugs.php.net/bug.php?id=77302

There are some workarounds and discussion about it : 

- https://github.com/symfony/symfony/issues/29459
- https://github.com/symfony/symfony/issues/30014
- https://bugs.php.net/bug.php?id=77302

For now I had to disable the profiler for the blocks. 
https://sonata-project.org/bundles/block/master/doc/reference/profiler.html

Thanks,
Jérôme.

toooni:
## Errors with deserialization for the profiling of blocks in debug mode with PHP 7.3

Instead of implementing an own solution for serialize/unserialize which resolves the PHP 7.3 bug https://bugs.php.net/bug.php?id=77302, I modified the `BlockDataCollector` to extend `Symfony\Component\HttpKernel\DataCollector\DateCollector`. This way we take advantage of the already solved serialize/unserialize issue there. 

I am targeting this branch, because this affects all users with PHP 7.3. This fix keeps BC even for already serialized data.

Closes #556

## Changelog

```markdown
### Changed
- Fix serializing issue for `BlockDataCollector`
```


src/Profiler/DataCollector/BlockDataCollector.php
```diff
@@ -16,45 +16,25 @@
...
 {
     /**
      * @var BlockHelper
      */
     protected $blocksHelper;
 
-    /**
-     * @var array
-     */
-    protected $blocks = [];
```
core23: Removing protected/public stuff is a BC break :(

toooni: I think this is ok since it's in a DataCollector. I know we should never assume what people do with a lib, but the worst that could happen is what happens at the moment with the issue ;) The profiler gives a error 500.

src/Profiler/DataCollector/BlockDataCollector.php
```diff
@@ -16,45 +16,25 @@
...
-use Symfony\Component\HttpKernel\DataCollector\DataCollectorInterface;
+use Symfony\Component\HttpKernel\DataCollector\DataCollector;
 
 /**
  * Block data collector for the symfony web profiling.
  *
  * @author Olivier Paradis <paradis.olivier@gmail.com>
  */
-class BlockDataCollector implements DataCollectorInterface, \Serializable
+class BlockDataCollector extends DataCollector
```
kunicmarko20: depending on a concrete class instead of an interface? why is this a good idea?

toooni: Because it's an abstract class provided by the framework which uses exactly this class.

phansys:
If there a way to add tests for this fix in order to avoid regressions?

core23:
Ping @toooni

toooni:
@core23 I've added an object in a test trace which should make a problem if we wouldn't use the `AbstractDataCollector` and PHP7.3. I don't think this doesn't help to prevent anything since - but I wouldn't know what else to do ;)

src/Templating/Helper/BlockHelper.php
```diff
@@ -284,7 +284,8 @@ public function render($block, array $options = [])
         }
 
         if ($this->stopwatch) {
-            $stats['cache']['created_at'] = $response->getDate();
+            // avoid \DateTime because of serialize/unserialize issue in PHP7.3 (https://bugs.php.net/bug.php?id=77302)
+            $stats['cache']['created_at'] = $response->getDate()->getTimestamp();
```
phansys: Is this change required for versions different than 7.3 too? I don't know how is it used, just asking to be sure if this change should modify the behavior for versions not affected by the issue too.
By other side, I see in the signature of `Response::getDate()` that it allows `null` as return value. Is that scenario not possible here?

toooni: I don't know but since the definition allows null I've fixed for null return values 👍

phansys: Thank you.
I've also shared another concern:
>Is this change required for versions different than 7.3 too? I don't know how is it used, just asking to be sure if this change should modify the behavior for versions not affected by the issue too.

toooni: Sorry. Wanted to check this first again. The change is not required except for PHP 7.3 - but it seems that the `created_at` var isn't used in the templates at all. So maybe the best solution would have been to remove this line.

phansys: Let's keep this as is then :+1:

phansys:
@toooni, could you please explain how to reproduce the issue this PR is fixing?
I'll try to check it locally.

core23:
Ping @toooni

toooni:
@phansys Here's a link to the issue (which is not associated with me) https://github.com/sonata-project/SonataBlockBundle/issues/556
Honestly - I can't reproduce it now. I don't know why and don't have the time right now to dig into it from the beginning (I've reproduced this in March). But relying on the `Abstract` `DataCollector` from Symfony reduces complexity (-39 lines of code) in this project and is still a good idea to avoid such issues in the future.

phansys:
I agree @toooni, thank you.
Could you please squash your commits?

toooni:
@phansys Can't you just choose "Squash and merge" in the merge drop down menu? Else I can do this next week.

greg0ire:
Thanks @toooni

PR #567, opened 2019-05-01
Removed admin-bundle dependency

core23:
## Subject

I am targeting this branch, because this contains BC breaks.

## Changelog

```markdown

### Changed
- Replace admin `FormMapper` with internal component

### Removed
- Removed `admin-bundle` dependency

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,6 @@
     "require-dev": {
         "jms/di-extra-bundle": "^1.7",
         "knplabs/knp-menu-bundle": "^2.0",
-        "sonata-project/admin-bundle": "^3.22 || ^4.0@dev",
         "symfony/debug": "^2.8 || ^3.2 || ^4.0",
         "symfony/phpunit-bridge": "^4.2",
         "symfony/stopwatch": "^2.8 || ^3.2 || ^4.0"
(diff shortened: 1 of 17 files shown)


kunicmarko20:
maybe port the test from admin bundle?

kunicmarko20:
(if there is any)

tests/Block/Service/RssBlockServiceTest.php
```diff
@@ -39,7 +40,7 @@ public function testService(): void
 
         $blockContext = new BlockContext($block, $optionResolver->resolve());
 
-        $formMapper = $this->getMockBuilder('Sonata\\AdminBundle\\Form\\FormMapper')
+        $formMapper = $this->getMockBuilder(FormMapper::class)
```
OskarStark: Can we use createMock here?

composer.json
```diff
@@ -35,6 +35,7 @@
         "symfony/http-foundation": "^2.8 || ^3.2 || ^4.0",
         "symfony/http-kernel": "^2.8 || ^3.2 || ^4.0",
         "symfony/options-resolver": "^2.8 || ^3.2 || ^4.0",
+        "symfony/templating": "^2.8 || ^3.2 || ^4.0",
```
kunicmarko20: this is deprecated, we should probably look into dropping it

PR #505, opened 2018-10-01
Added "debug:block" command alias

core23:
I am targeting this branch, because this is BC.

## Changelog

```markdown
### Added
- Added `debug:sonata:block` command alias for `DebugBlocksCommand`
```

## Subject

You should use the `debug:` prefix for debug commands.

PR #574, opened 2019-06-21
Stop extending deprecated class `ContainerAwareCommand`

phansys:
## Subject

Stop extending deprecated class `ContainerAwareCommand`.

I am targeting this branch, because these changes tries to respect BC as much as possible.

## Changelog

```markdown
### Deprecated
- `BaseCommand::getBlockServiceManager()` method in favor of `BaseCommand::$blockManager` property;
- Extending `DebugBlocksCommand` class, which will be declared final in 4.0;
- Invoking `DebugBlocksCommand` with "debug:sonata:block" as name.

### Fixed
- Deprecation caused by usage of `ContainerAwareCommand`.
```
    
## To do
    
- [x] Add tests;
- [x] Validate with core members why the command defined by `DebugBlocksCommand` requires an alias, and check if we must instead deprecate invocations with some of these names.

### DIFF
--- a/src/Command/BaseCommand.php
+++ b/src/Command/BaseCommand.php
@@ -14,15 +14,46 @@ declare(strict_types=1);
 namespace Sonata\BlockBundle\Command;
 
 use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
-use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Command\Command;
 
-abstract class BaseCommand extends ContainerAwareCommand
+abstract class BaseCommand extends Command
 {
+    /**
+     * @var BlockServiceManagerInterface
+     */
+    protected $blockManager;
+
+    public function __construct(string $name = null, BlockServiceManagerInterface $blockManager = null)
+    {
+        // NEXT_MAJOR: Remove the default value for argument 2 and the following condition
+        if (null === $blockManager) {
+            throw new \InvalidArgumentException(sprintf(
+                'Argument 2 passed to %s::%s() must be an instance of %s, %s given.',
+                static::class,
+                __FUNCTION__,
+                BlockServiceManagerInterface::class,
+                \gettype($blockManager)
+            ));
+        }
+
+        $this->blockManager = $blockManager;
+
+        parent::__construct($name);
+    }
+
     /**
      * @return BlockServiceManagerInterface
      */
     public function getBlockServiceManager()
     {
-        return $this->getContainer()->get('sonata.block.manager');
+        // NEXT_MAJOR: Remove this method
+        @trigger_error(sprintf(
+            'Method %1$s::%2$s() is deprecated since sonata-project/block-bundle 3.x and will be removed with the 4.0 release.'.
+            'Use the %1$s::$blockManager property instead.',
+            static::class,
+            __FUNCTION__
+        ), E_USER_DEPRECATED);
+
+        return $this->blockManager;
     }
 }
--- a/src/Command/DebugBlocksCommand.php
+++ b/src/Command/DebugBlocksCommand.php
@@ -18,16 +18,28 @@ use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
-class DebugBlocksCommand extends BaseCommand
+/**
+ * @final since sonata-project/block-bundle 4.0
+ *
+ * NEXT_MAJOR: Uncomment the "final" class declaration
+ */
+/* final */class DebugBlocksCommand extends BaseCommand
 {
+    /**
+     * {@inheritdoc}
+     *
+     * NEXT_MAJOR: Rename to "debug:sonata:block"
+     */
+    protected static $defaultName = 'sonata:block:debug';
+
     /**
      * {@inheritdoc}
      */
     public function configure()
     {
-        // NEXT_MAJOR: Switch name and alias
+        $this->setName(static::$defaultName); // BC for symfony/console < 3.4.0
+        // NEXT_MAJOR: Replace the current alias by "sonata:block:debug"
         $this->setAliases(['debug:sonata:block']);
-        $this->setName('sonata:block:debug');
         $this->setDescription('Debug all blocks available, show default settings of each block');
 
         $this->addOption('context', 'c', InputOption::VALUE_REQUIRED, 'display service for the specified context');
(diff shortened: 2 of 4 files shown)


src/Command/BaseCommand.php
```diff
@@ -14,15 +14,41 @@
...
     public function getBlockServiceManager()
     {
-        return $this->getContainer()->get('sonata.block.manager');
+        // NEXT_MAJOR: Remove this method
+        @trigger_error(sprintf(
+            'Method %1$s::%2$s() is deprecated since 3.x and will be removed with the 4.0 release.'.
+            'Use the %1$s::$blockManager property instead.',
+            static::class,
+            __FUNCTION__
+        ),
```
greg0ire:
```suggestion
        ), E_USER_DEPRECATED);
```

src/Command/BaseCommand.php
```diff
@@ -14,15 +14,41 @@
...
+        E_USER_DEPRECATED);
```
greg0ire:
```suggestion
```

src/Command/BaseCommand.php
```diff
@@ -14,15 +14,41 @@
...
+
     /**
      * @return BlockServiceManagerInterface
      */
     public function getBlockServiceManager()
     {
-        return $this->getContainer()->get('sonata.block.manager');
+        // NEXT_MAJOR: Remove this method
+        @trigger_error(sprintf(
+            'Method %1$s::%2$s() is deprecated since 3.x and will be removed with the 4.0 release.'.
```
greg0ire: When you deprecate something, please include the package name :pray: 

```suggestion
            'Method %1$s::%2$s() is deprecated since sonata-project/block-bundle 3.x and will be removed with the 4.0 release.'.
```

greg0ire:
```suggestion
            'Method %s is deprecated since 3.x and will be removed with the 4.0 release.'.
```

src/Command/BaseCommand.php
```diff
@@ -14,15 +14,41 @@
...
+            'Use the %1$s::$blockManager property instead.',
+            static::class,
```
greg0ire:
```suggestion
            __METHOD__
```

src/Command/BaseCommand.php
```diff
@@ -14,15 +14,41 @@
...
+            __FUNCTION__
```
greg0ire:
```suggestion
```

src/Command/DebugBlocksCommand.php
```diff
@@ -18,16 +18,28 @@
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
-class DebugBlocksCommand extends BaseCommand
+/**
+ * @final since 4.0
+ *
+ * NEXT_MAJOR: Comment out the "final" class declaration
```
greg0ire: I think you mean the opposite :wink: 

https://english.stackexchange.com/questions/33483/when-i-say-comment-out-does-it-mean-to-uncomment-something-or-comment-it/33502
```suggestion
 * NEXT_MAJOR: Uncomment the "final" class declaration
```

tests/Command/DebugBlocksCommandTest.php
```diff
@@ -0,0 +1,93 @@
...
+    {
+        $this->application = new Application();
+
+        $blockManager = $this->createMock(BlockServiceManagerInterface::class);
+        $blockManager
+            ->expects($this->any())
+            ->method('getServices')
+            ->willReturn([]);
+
+        $command = new DebugBlocksCommand(null, $blockManager);
```
greg0ire:
```suggestion
        $this->application->add(new DebugBlocksCommand(null, $blockManager));
```

tests/Command/DebugBlocksCommandTest.php
```diff
@@ -0,0 +1,93 @@
...
+
+        $this->application->add($command);
```
greg0ire:
```suggestion
```

tests/Command/DebugBlocksCommandTest.php
```diff
@@ -0,0 +1,93 @@
...
+        $this->application = new Application();
+
+        $blockManager = $this->createMock(BlockServiceManagerInterface::class);
+        $blockManager
+            ->expects($this->any())
+            ->method('getServices')
+            ->willReturn([]);
+
+        $command = new DebugBlocksCommand(null, $blockManager);
+
```
greg0ire:
```suggestion
```

src/Command/BaseCommand.php
phansys: If we use method here, the deprecation message will refer to the abstract class instead of the called class. Is that fine?

phansys: Just for context, this is similar to what is used [here](https://github.com/sonata-project/SonataAdminBundle/blob/64b83e290566dc609a7a30d6a419c1359430dfc9/src/Admin/AbstractAdmin.php#L917) to show the instantiated class name instead the abstracted one.

greg0ire: Oh right, nvm then, keep it that way please.

phansys:
Sure @greg0ire. I tend to squash when there is some consensus about the PR status, in order to not loose the discussion. Thank you so much for your hints.

phansys:
Before merging this PR, I'd like to resolve the *TO DO*. Do you know why we need an alias for this command?

greg0ire:
A while ago, symfony migrated the `container:debug` and `router:debug` commands to `debug:container` and `debug:router`. I think the alias is here to provide consistency with that `debug:` prefix, and that we should only have the `debug:` version work on next major. See https://github.com/sonata-project/SonataBlockBundle/pull/505

phansys:
Should we then deprecate the invocations using the old name?

greg0ire:
Yes, I think we should :)

phansys:
Deprecation added for old command invocation :+1:

src/Command/BaseCommand.php
```diff
@@ -14,15 +14,40 @@
...
-use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Command\Command;
 
-abstract class BaseCommand extends ContainerAwareCommand
+abstract class BaseCommand extends Command
 {
+    /**
+     * @var BlockServiceManagerInterface
+     */
+    protected $blockManager;
```
greg0ire: Why not `private`?

src/Command/BaseCommand.php
```diff
@@ -14,15 +14,40 @@
...
+
+    public function __construct(string $name = null, BlockServiceManagerInterface $blockManager = null)
+    {
+        // NEXT_MAJOR: Remove the default value for argument 2 and the following condition
+        if (null === $blockManager) {
+            throw new \InvalidArgumentException(sprintf('Argument 2 passed to %s::%s() must be an instance of %s, %s given.', static::class, __FUNCTION__, BlockServiceManagerInterface::class, \gettype($blockManager)));
```
greg0ire: Please use linebreaks

src/Command/DebugBlocksCommand.php
```diff
@@ -38,10 +50,17 @@ public function configure()
      */
     public function execute(InputInterface $input, OutputInterface $output)
     {
+        if ('sonata:block:debug' === $input->getArgument('command')) {
+            // NEXT_MAJOR: Remove this check
+            @trigger_error(
+                'Command "sonata:block:debug" is deprecated since sonata-project/block-bundle 3.x and will be removed with the 4.0 release.'.
+                ' Use the "debug:sonata:block" command instead.',
+                E_USER_DEPRECATED);
```
greg0ire: From [PSR-2](https://www.php-fig.org/psr/psr-2/):

> Argument lists MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one argument per line.

src/Command/BaseCommand.php
phansys: Because this property must be accessible from `DebugBlocksCommand`.

src/Command/BaseCommand.php
phansys: Done.

PR #575, opened 2019-06-21
[Tests] Fix some test file paths

phansys:
## Subject

Fix some test file paths.

I am targeting this branch, because these changes are BC.

### DIFF
--- a/tests/Block/Service/AbstractBlockServiceTestCase.php
+++ b/tests/Block/Service/AbstractBlockServiceTestCase.php
@@ -11,7 +11,7 @@ declare(strict_types=1);
  * file that was distributed with this source code.
  */
 
-namespace Sonata\BlockBundle\Test;
+namespace Sonata\BlockBundle\Tests\Block\Service;
 
 use PHPUnit\Framework\TestCase;
 use Sonata\BlockBundle\Block\BlockContextInterface;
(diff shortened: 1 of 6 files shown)


tests/Block/Service/AbstractBlockServiceTestCase.php
```diff
@@ -11,14 +11,15 @@
  * file that was distributed with this source code.
  */
 
-namespace Sonata\BlockBundle\Test;
+namespace Sonata\BlockBundle\Tests\Block\Service;
```
kunicmarko20: I think the idea here is to make it reusable for other libraries and that is why it is in src and not tests

core23:
This is not a bug, this is intended to reuse the classes in other projects

phansys:
I understand, although I think it's a non common or standardised practice.
Is this intention focused to CI purposes in order to allow other packages of this project to leverage these classes, or to any project?

kunicmarko20:
Its common practice, https://github.com/symfony/symfony/tree/4.4/src/Symfony/Bundle/FrameworkBundle/Test

phansys:
Well, this seems reasonable to me then. I didn't see https://sonata-project.org/bundles/block/master/doc/reference/testing.html#test-blocks before.
Thank you.

PR #576, opened 2019-06-22
Add as many type hints as possible

greg0ire:
## Subject

This is technically a BC-break, but these interfaces have no implementations
yet.

I'm not sure why the type hints were not added in the first place.
Next step would be to implement these in the admin bundle.

### DIFF
--- a/src/Block/Service/EditableBlockService.php
+++ b/src/Block/Service/EditableBlockService.php
@@ -29,10 +29,5 @@ interface EditableBlockService
 
     public function validate(ErrorElement $errorElement, BlockInterface $block);
 
-    /**
-     * @param string|null $code
-     *
-     * @return MetadataInterface
-     */
-    public function getMetadata($code = null);
+    public function getMetadata(): MetadataInterface;
 }
--- a/src/Form/Mapper/FormMapper.php
+++ b/src/Form/Mapper/FormMapper.php
@@ -21,56 +21,30 @@ use Symfony\Component\Form\FormBuilderInterface;
 interface FormMapper
 {
     /**
-     * @param string $name
-     * @param mixed  $type
-     *
-     * @return FormBuilderInterface
+     * @param mixed $type
      */
-    public function create($name, $type = null, array $options = []);
+    public function create(string $name, ?string $type = null, array $options = []): FormBuilderInterface;
 
     /**
-     * @param array $keys field names
-     *
-     * @return self
+     * @param string[] $keys field names
      */
-    public function reorder(array $keys);
+    public function reorder(array $keys): self;
 
     /**
      * @param FormBuilderInterface|string $name
-     * @param string                      $type
-     *
-     * @return self
      */
-    public function add($name, $type = null, array $options = []);
+    public function add($name, ?string $type = null, array $options = []): self;
+
+    public function remove(string $key): self;
+
+    public function setHelps(array $helps = []): self;
+
+    public function addHelp(string $name, string $help): self;
+
+    public function has(string $key): bool;
 
     /**
-     * @param string $key
-     *
-     * @return self
-     */
-    public function remove($key);
-
-    /**
-     * @return self
-     */
-    public function setHelps(array $helps = []);
-
-    /**
-     * @return self
-     */
-    public function addHelp($name, $help);
-
-    /**
-     * @param string $key
-     *
-     * @return bool
-     */
-    public function has($key);
-
-    /**
-     * @param string $key
-     *
      * @return mixed
      */
-    public function get($name);
+    public function get(string $name);
 }


src/Block/Service/EditableBlockService.php
```diff
@@ -29,10 +29,5 @@ public function configureCreateForm(FormMapper $form, BlockInterface $block);
 
     public function validate(ErrorElement $errorElement, BlockInterface $block);
 
-    /**
-     * @param string|null $code
-     *
-     * @return MetadataInterface
-     */
-    public function getMetadata($code = null);
+    public function getMetadata(?string $code = null): MetadataInterface;
```
phansys: Is the `?` modifier required if this argument already supports `null` as value via its default value?

greg0ire: It is, because implementors might use `string $code = ''`, and that would make it no longer `nullable`. Now they are forced to support `null`

src/Form/Mapper/FormMapper.php
```diff
@@ -21,56 +21,30 @@
...
 {
     /**
-     * @param string $name
-     * @param mixed  $type
-     *
-     * @return FormBuilderInterface
+     * @param mixed $type
      */
-    public function create($name, $type = null, array $options = []);
+    public function create(string $name, $type = null, array $options = []): FormBuilderInterface;
```
core23:
```suggestion
    public function create(string $name, ?string $type = null, array $options = []): FormBuilderInterface;
```
It make no sense to pass an object here. It will call symfony `FormBuilder::create`
https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Form/FormBuilderInterface.php#L34-L43

src/Block/Service/EditableBlockService.php
core23: I'm not sure if we even need this parameter...

We use this parameter as a fallback for the descriptions which looks wrong
https://github.com/sonata-project/SonataNewsBundle/blob/3.x/src/Block/RecentPostsBlockService.php#L152-L157

core23: We should deprecate this, even if we have no running implementation of this interface

greg0ire: Should I just remove it? It's still compatible with implementations that have a this be a default argument: https://3v4l.org/lIYFt

core23: Yes, please remove this 👍

phansys:
It is still experimental, but maybe we should give a try to **phpdoc_to_return_type** PHP CS fixer rule: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3640.

greg0ire:
I don't think we should, adding return types can be a BC-break in some situations… or maybe only on master branches?

src/Form/Mapper/FormMapper.php
```diff
@@ -21,56 +21,30 @@
...
 {
     /**
-     * @param string $name
-     * @param mixed  $type
-     *
-     * @return FormBuilderInterface
+     * @param mixed $type
      */
-    public function create($name, $type = null, array $options = []);
+    public function create(string $name, ?string $type = null, array $options = []): FormBuilderInterface;
```
phansys: Except of what I suggested before about the 2nd nullable argument, the other changes are fine on my side.

greg0ire: Sorry, aren't you happy with the justification I gave?

phansys:
Yes, I mean to test that rule on `master` in order to know if it eases our process of typing return declarations. If that provides good results, maybe we should commit the changes but keeping the rule out of the scope. At least, until that will be stable.

src/Form/Mapper/FormMapper.php
phansys: Didn't see the last comments, sorry. I'm fine with that then. I'm pretty dizzy with the LSP implementation changes ([covariants](https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters) and co, etc). I have a vague memory about this syntax being discouraged when the argument already defines `null` as default value, but I can't confirm that now.

src/Form/Mapper/FormMapper.php
```diff
@@ -21,56 +21,30 @@
...
-     * @return self
-     */
-    public function setHelps(array $helps = []);
+    public function setHelps(array $helps = []): self;
 
-    /**
-     * @return self
-     */
-    public function addHelp($name, $help);
+    public function addHelp($name, $help): self;
```
phansys: I don't know any implementation of this interface, but I'm pretty sure these arguments are always strings. At least is what happen at [`Sonata\AdminBundle\Form\FormMapper::addHelp()`](https://github.com/sonata-project/SonataAdminBundle/blob/bf184d36a26fe47e6b0db2c8e18d03bbd9cbcab4/src/Form/FormMapper.php#L256).
```suggestion
    public function addHelp(string $name, string $help): self;
```

greg0ire: @core23 you wrote that interface, what do you think? Sounds reasonable…

core23: They are always string :+1:

Issue #585, opened 2019-07-05
Remove implicit dependency against SonataAdminBundle


phansys:
## Feature Request
Remove implicit dependency against SonataAdminBundle by removing all the usages of classes provided by `Sonata\AdminBundle\` namespace.

This is a follow up of sonata-project/SonataAdminBundle#5622.

https://github.com/sonata-project/SonataBlockBundle/blob/d0a67fcd6873e234467cd58b35aa3feb5b54a6ff/src/Block/Service/AbstractAdminBlockService.php#L16
https://github.com/sonata-project/SonataBlockBundle/blob/d0a67fcd6873e234467cd58b35aa3feb5b54a6ff/src/Event/TextBlockListener.php#L16

### To do
* Declarations of `Sonata\AdminBundle\Form\FormMapper` must be replaced with `Sonata\BlockBundle\Form\Mapper\FormMapper`;
* Find a replacement or a way to avoid the usage of `Sonata\AdminBundle\Admin\AdminInterface`.

core23:
Already done on the master branch

PR #586, opened 2019-07-05
Remove implicit depencency against "sonata-project/admin-bundle"

phansys:
## Subject

Remove implicit dependency against "sonata-project/admin-bundle".

Triggered by #585.

## Changelog

```markdown
### Changed
- Replaced declarations of `Sonata\AdminBundle\Form\FormMapper` with `Sonata\BlockBundle\Form\Mapper\FormMapper`.
```

## To do
    
- [ ] Wait for the release of sonata-project/SonataAdminBundle#5623 in order to ensure `Sonata\AdminBundle\Form\FormMapper` is implementing `Sonata\BlockBundle\Form\Mapper\FormMapper`;
- [ ] Find a replacement or another way to avoid the usage of `Sonata\AdminBundle\Admin\AdminInterface`.

### DIFF
--- a/src/Block/Service/AbstractAdminBlockService.php
+++ b/src/Block/Service/AbstractAdminBlockService.php
@@ -13,7 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\BlockBundle\Block\Service;
 
-use Sonata\AdminBundle\Form\FormMapper;
+use Sonata\BlockBundle\Form\Mapper\FormMapper;
 use Sonata\BlockBundle\Meta\Metadata;
 use Sonata\BlockBundle\Model\BlockInterface;
 use Sonata\CoreBundle\Validator\ErrorElement;
--- a/src/Block/Service/AdminBlockServiceInterface.php
+++ b/src/Block/Service/AdminBlockServiceInterface.php
@@ -13,7 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\BlockBundle\Block\Service;
 
-use Sonata\AdminBundle\Form\FormMapper;
+use Sonata\BlockBundle\Form\Mapper\FormMapper;
 use Sonata\BlockBundle\Meta\MetadataInterface;
 use Sonata\BlockBundle\Model\BlockInterface;
 use Sonata\CoreBundle\Validator\ErrorElement;
--- a/src/Block/Service/ContainerBlockService.php
+++ b/src/Block/Service/ContainerBlockService.php
@@ -13,8 +13,8 @@ declare(strict_types=1);
 
 namespace Sonata\BlockBundle\Block\Service;
 
-use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\BlockBundle\Block\BlockContextInterface;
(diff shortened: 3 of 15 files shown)


core23:
You can't do this on a stable branch, AFAIK I already did this for the master branch

phansys:
I understand this could be technically a BC break, but just in case the user is checking against `get_declared_interfaces()` or similar.
Excluding the case of `Sonata\AdminBundle\Admin\AdminInterface`, my reasoning is the following:

**SonataAdminBundle**:
Let's suppose that the next stable release of SonataAdminBundle implements the `FormInterface` from SonataBlockBundle. There is no problem there since there are no changes in the SonataAdminBundle's API: all the methods from the new interface were already implemented.

**SonataBlockBundle**:
Now, the next stable release of SonataBlockBundle is able to remove the declarations against the implementations on SonataAdminBundle, since if we add a conflict with the range of versions that don't implement the new interfaces, we can be sure there will be no problem when the user performs an update.
This way, we will block new updates from SonataBlockBundle if the user doesn't update SonataAdminBundle to the required version too.

Maybe there is something I can be missing, but please let me know if you have some comments or considerations with this approach.

Thank you!

core23:
The problem is, that we have an interface change which will cause some signature problems across the bundles and custom block services:
https://github.com/sonata-project/SonataBlockBundle/blob/bcd407d2c9f73cab07256824468f6d3859064e4e/src/Block/Service/AdminBlockServiceInterface.php#L27-L42
The `FormMapper` is a different class now. It's a technical BC break which is caused by PHP itself.

phansys:
>The FormMapper is a different class now. It's a technical BC break which is caused by PHP itself.

Let's recreate the BC break scenario I guess you're proposing:

* The user has their own implementation extending `Sonata\AdminBundle\Form\FormMapper`, which is part of the contract of some methods at `AdminBlockServiceInterface`.
* The user tries to update SonataBlockBundle to the next version which breaks the contract, and composer blocks the update until the user updates SonataAdminBundle to the next version that ships the new interface implementation, which is the base for its custom implementation.
* The user updates SonataAdminBundle to the next version with ships the interface implementation from SonataBlockBundle, and the contracts from `AdminBlockServiceInterface` remain respected.

AFAIK, with the update path I'm proposing, there should not be any issue, since the class passed to the contracts proposed by `AdminBlockServiceInterface` should remain the same type.

core23:
I'm speaking of a different scenario. This should be a test case which explains the problem:
https://3v4l.org/ifFkc

We changed the method signature of the `AdminBlockServiceInterface` methods by providing a different parameter type. This should be no problem, as we provide an implementation (`AdminFormMapper`) of the same interface (`BlockFormMapper`), but this will break the PHP compiler.

We can adopt (or add a conflict entry to the `composer.json` file) our projects which are using the BlockBundle, but not the projects on the user site.

phansys:
I see, thank you for the explanation. We strongly need covariants support.
So, there is no way to do this change on 3.x.

core23:
> So, there is no way to do this change on 3.x.

Sadly not. That's the reason why we should release the 4.0 release asap. So we can remove some old legacy stuff.

PR #587, opened 2019-07-06
Deprecated AbstractBlockServiceTestCase class with BlockServiceTestCase

### LINKED ISSUE #425: AbstractBlockServiceTestCase should be renamed to BlockServiceTestCase

core23:
## Subject

This class is loaded inside the prod autoloader, we should handle this as a `minor` change.

Closes #425

## Changelog

```markdown
### Deprecated
- Deprecated `AbstractBlockServiceTestCase` class with `BlockServiceTestCase`
```

### DIFF
--- a/src/Test/AbstractBlockServiceTestCase.php
+++ b/src/Test/AbstractBlockServiceTestCase.php
@@ -13,91 +13,20 @@ declare(strict_types=1);
 
 namespace Sonata\BlockBundle\Test;
 
-use PHPUnit\Framework\TestCase;
-use Sonata\BlockBundle\Block\BlockContextInterface;
-use Sonata\BlockBundle\Block\BlockContextManager;
-use Sonata\BlockBundle\Block\BlockContextManagerInterface;
-use Sonata\BlockBundle\Block\BlockServiceInterface;
-use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
+@trigger_error(
+    'The '.__NAMESPACE__.'\AbstractBlockServiceTestCase class is deprecated since sonata-project/block-bundle 3.x '.
+    'and will be removed with the 4.0 release. '.
+    'Use '.__NAMESPACE__.'\BlockServiceTestCase instead.',
+    E_USER_DEPRECATED
+);
 
 /**
  * Abstract test class for block service tests.
  *
  * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ *
+ * @deprecated since sonata-project/block-bundle 3.x, to be removed in 4.0. Use Sonata\BlockBundle\Test\BlockServiceTestCase instead.
  */
-abstract class AbstractBlockServiceTestCase extends TestCase
+abstract class AbstractBlockServiceTestCase extends BlockServiceTestCase
 {
-    /**
-     * @var \PHPUnit_Framework_MockObject_MockObject|ContainerInterface
-     */
-    protected $container;
-
-    /**
-     * @var \PHPUnit_Framework_MockObject_MockObject|BlockServiceManagerInterface
-     */
-    protected $blockServiceManager;
-
-    /**
-     * @var BlockContextManagerInterface
-     */
-    protected $blockContextManager;
-
-    /**
-     * @var FakeTemplating
-     */
-    protected $templating;
-
-    protected function setUp()
-    {
-        $this->container = $this->createMock('Symfony\Component\DependencyInjection\ContainerInterface');
-        $this->templating = new FakeTemplating();
-
-        $blockLoader = $this->createMock('Sonata\BlockBundle\Block\BlockLoaderInterface');
-        $this->blockServiceManager = $this->createMock('Sonata\BlockBundle\Block\BlockServiceManagerInterface');
-        $this->blockContextManager = new BlockContextManager($blockLoader, $this->blockServiceManager);
-    }
-
-    /**
-     * Create a mocked block service.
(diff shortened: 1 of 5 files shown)


src/Test/AbstractBlockServiceTestCase.php
```diff
@@ -13,91 +13,20 @@
...
 
-use PHPUnit\Framework\TestCase;
-use Sonata\BlockBundle\Block\BlockContextInterface;
-use Sonata\BlockBundle\Block\BlockContextManager;
-use Sonata\BlockBundle\Block\BlockContextManagerInterface;
-use Sonata\BlockBundle\Block\BlockServiceInterface;
-use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
+@trigger_error(
+    'The '.__NAMESPACE__.'\AbstractBlockServiceTestCase class is deprecated since 3.x '.
```
phansys:
```suggestion
    'The '.__NAMESPACE__.'\AbstractBlockServiceTestCase class is deprecated since sonata-project/block-bundle 3.x '.
```

src/Test/BlockServiceTestCase.php
```diff
@@ -0,0 +1,103 @@
...
+     * @var BlockContextManagerInterface
+     */
+    protected $blockContextManager;
+
+    /**
+     * @var FakeTemplating
+     */
+    protected $templating;
+
+    protected function setUp()
```
phansys:
```suggestion
    protected function setUp(): void
```

core23: We can't do this. This will not work in PHP 7.1 :(

src/Test/BlockServiceTestCase.php
```diff
@@ -0,0 +1,103 @@
...
+
+        $blockLoader = $this->createMock(BlockLoaderInterface::class);
+        $this->blockServiceManager = $this->createMock(BlockServiceManagerInterface::class);
+        $this->blockContextManager = new BlockContextManager($blockLoader, $this->blockServiceManager);
+    }
+
+    /**
+     * Create a mocked block service.
+     *
+     * @param BlockServiceInterface $blockService A block service
```
greg0ire: I think we can remove this one

src/Test/BlockServiceTestCase.php
```diff
@@ -0,0 +1,103 @@
...
+     */
+    protected function getBlockContext(BlockServiceInterface $blockService): BlockContextInterface
```
greg0ire: Should we use `final` here and below?

core23: We can't do this on the stable branch, also this are just test extension points :/

greg0ire: It's not a BC-break: it's a new class, people are supposed to change their code to migrate anyway. But if those are extension points, then ok.

core23:
Can you please review and merge @phansys ?

src/Test/BlockServiceTestCase.php
phansys: I don't understand why this will not work. Could you please explain?

core23: `void` needs PHP 7.1
https://wiki.php.net/rfc/void_return_type

phansys: Sure, so "This will not work in PHP < 7.1"
It's now clarified to me :+1:

phansys: Currently, we're requiring `7.1` as minimum version:
https://github.com/sonata-project/SonataBlockBundle/blob/d0a67fcd6873e234467cd58b35aa3feb5b54a6ff/composer.json#L23

core23: Oh sorry, I was adressing the wrong problem. The problem is PHP 7.1 itself:
```
Fatal error: Declaration of Sonata\BlockBundle\Tests\Block\Service\MenuBlockServiceTest::setUp() must be compatible with Sonata\BlockBundle\Test\BlockServiceTestCase::setUp(): void in /app/tests/Block/Service/MenuBlockServiceTest.php on line 26
```
You can't override the method signature in PHP 7.1. The PHPUnit method signature was changed:
https://github.com/sebastianbergmann/phpunit/blob/7.5/src/Framework/TestCase.php#L407
https://github.com/sebastianbergmann/phpunit/blob/8.2/src/Framework/TestCase.php#L411

phansys:
>Can you please review and merge @phansys ?

I've approved the PR, but I can't merge because the branch protection:
```
 ! [remote rejected] 3x_merging -> 3.x (protected branch hook declined)
error: failed to push some refs to 'git@github.com:sonata-project/SonataBlockBundle.git'
```

core23:
> > Can you please review and merge @phansys ?
> 
> I've approved the PR, but I can't merge because the branch protection:
> 
> ```
>  ! [remote rejected] 3x_merging -> 3.x (protected branch hook declined)
> error: failed to push some refs to 'git@github.com:sonata-project/SonataBlockBundle.git'
> ```

Did you use the Web UI for the merge?

phansys:
No, I usually merge from CLI. Should I use the web UI for this project?

core23:
Never used the CLI for GitHub PRs

PR #591, opened 2019-07-06
Use symfony contracts

core23:
## Subject

I am targeting this branch, because this is a BC break, 'cause we changed the parent class of `BlockEvent`.

## Changelog

```markdown
### Changed
- Use symfony event dispatcher contracts 
### Removed
- Removed `symfony/event-dispatcher` dependency
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,6 @@
         "symfony/config": "^2.8 || ^3.2 || ^4.0",
         "symfony/console": "^2.8 || ^3.2 || ^4.0",
         "symfony/dependency-injection": "^2.8 || ^3.2 || ^4.0",
-        "symfony/event-dispatcher": "^2.8 || ^3.2 || ^4.0",
         "symfony/event-dispatcher-contracts": "^1.1",
         "symfony/form": "^2.8 || ^3.2 || ^4.0",
         "symfony/framework-bundle": "^2.8 || ^3.2 || ^4.0",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -31,6 +31,7 @@
         "symfony/console": "^2.8 || ^3.2 || ^4.0",
         "symfony/dependency-injection": "^2.8 || ^3.2 || ^4.0",
         "symfony/event-dispatcher": "^2.8 || ^3.2 || ^4.0",
```
phansys: With these changes, there are no usages left from `Symfony\Component\EventDispatcher\` namespace.
Should not we remove the dependency too?

core23: Not sure about this. We would remove an implementation for this thing: /

greg0ire: I think people using it should require it. IMO we should remove it.

composer.json
```diff
@@ -30,7 +30,7 @@
         "symfony/config": "^2.8 || ^3.2 || ^4.0",
         "symfony/console": "^2.8 || ^3.2 || ^4.0",
         "symfony/dependency-injection": "^2.8 || ^3.2 || ^4.0",
-        "symfony/event-dispatcher": "^2.8 || ^3.2 || ^4.0",
```
OskarStark: Wouldn’t this drop support for SF 2 and 3 ?

core23: Refs #592 592

PR #599, opened 2019-07-10
DevKit updates for master branch

SonataCI:


src/Block/BlockContextManager.php
```diff
@@ -54,12 +54,6 @@ class BlockContextManager implements BlockContextManagerInterface
...
     protected $logger;
 
-    /**
-     * @param BlockLoaderInterface         $blockLoader
-     * @param BlockServiceManagerInterface $blockService
-     * @param array                        $cacheBlocks
-     * @param LoggerInterface|null         $logger
-     */
     public function __construct(BlockLoaderInterface $blockLoader, BlockServiceManagerInterface $blockService,
         array $cacheBlocks = [], LoggerInterface $logger = null
```
OskarStark: this linebreaking looks weird and I think even @greg0ire is 👎 with that 😄

phansys: Maybe we should configure **method_argument_space.on_multiline** rule at DevKit.

greg0ire: Yeah I'm very much against it, to the point I contributed something in php-cs-fixer to prevent that. It was then transformed in the rule @phansys is talking about, which isn't enabled by default, probably for BC reasons

OskarStark: can you create a PR on dev-kit @phansys ?

phansys: Sure, I'll add the rule.

PR #602, opened 2019-07-12
Cleanup for major version

### LINKED ISSUE #609: Use explicit API
## Feature Request

- Make the API as explicit as possible with typehints
- Close all classes
- Prefer `private` visibility

core23:
## Subject

Fixes #609

I am targeting this branch, because this is a BC break.

No changelog, because this removes all NEXT_MAJOR hacks which are already mentioned by the general deprecation note inside the UPRADE-4.0.md file.


tests/DependencyInjection/Compiler/TweakCompilerPassTest.php
```diff
@@ -80,13 +80,19 @@ public function testProcessSameBlockId(): void
         $pass->process($this->container);
     }
 
+    public function blockIds(): \Generator
```
OskarStark:
```suggestion
    public function validBlockIds(): \Generator
```

?

core23:
Ready for review! @sonata-project/contributors

src/Block/Service/MenuBlockService.php
```diff
@@ -179,9 +181,10 @@ protected function getFormSettingsKeys(): array
      * Gets the menu to render.
      *
      *
+     *
```
phansys: These extra line LFs could be removed.

core23: Fixed

core23:
Two approvements, so... Let's merge this thin 🙈

PR #604, opened 2019-07-13
Deprecated BaseCommand

core23:
## Subject

This abstract class in only used once.

## Changelog

```markdown
### Deprecated
- Deprecated `BaseCommand` class

```

### DIFF
--- a/src/Command/BaseCommand.php
+++ b/src/Command/BaseCommand.php
@@ -16,6 +16,9 @@ namespace Sonata\BlockBundle\Command;
 use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
 use Symfony\Component\Console\Command\Command;
 
+/**
+ * @deprecated since sonata-project/block-bundle 3.x, to be removed in 4.0
+ */
 abstract class BaseCommand extends Command
 {
     /**


core23:
> Shouldn't we add the corresponding `trigger_error()` call with the `E_USER_DEPRECATED` as the error type before the class declaration?

This would trigger every time, because this class is used by `DebugBlocksCommand`

phansys:
Aren't these example cases resulting in the same behavior?
https://github.com/sonata-project/SonataBlockBundle/blob/d0a67fcd6873e234467cd58b35aa3feb5b54a6ff/src/Block/AbstractBlockService.php#L16-L21

https://github.com/sonata-project/SonataBlockBundle/blob/d0a67fcd6873e234467cd58b35aa3feb5b54a6ff/src/Block/BaseBlockService.php#L18-L23

core23:
> Aren't these example cases resulting in the same behavior?
> 
> https://github.com/sonata-project/SonataBlockBundle/blob/d0a67fcd6873e234467cd58b35aa3feb5b54a6ff/src/Block/AbstractBlockService.php#L16-L21
> 
> https://github.com/sonata-project/SonataBlockBundle/blob/d0a67fcd6873e234467cd58b35aa3feb5b54a6ff/src/Block/BaseBlockService.php#L18-L23

Yes, you're right. We should fix this somehow :|

phansys:
Maybe I'm missing something, but AFAIK these errors are only triggered when the file is loaded, and thanks the autoloading it occurs just once.

core23:
> Maybe I'm missing something, but AFAIK these errors are only triggered when the file is loaded, and thanks the autoloading it occurs just once.

That's right, but the user gets a deprecation warning without doing anything wrong

PR #606, opened 2019-07-20
Merge branch '3.x' into master

core23:


src/DependencyInjection/Compiler/TweakCompilerPass.php
```diff
@@ -51,7 +51,11 @@ public function process(ContainerBuilder $container): void
         }
 
         foreach ($container->findTaggedServiceIds('knp_menu.menu') as $serviceId => $tags) {
+            foreach ($tags as $tags) {
             foreach ($tags as $attributes) {
+                if (empty($attributes['alias'])) {
```
phansys: This check is duplicated and I think `$id` is not defined here.

core23:
There's something wrong with this PR 😕

PR #607, opened 2019-07-20
Merge branch '3.x' into master

core23:


src/Block/Service/AbstractBlockService.php
```diff
@@ -35,10 +35,31 @@ abstract class AbstractBlockService implements BlockServiceInterface
      */
     private $twig;
 
-    public function __construct(string $name, Environment $twig)
+    /**
```
phansys: These changes seem wrong.

core23: Why? We changed stuff in the stable and master branch:
**stable**: We deprecated the name
**master**: We added type hints

Removing the deprecated stuff should be done afterwards

phansys: In that case, it's fine for me.

PR #612, opened 2019-07-25
Remove some superfluous docblocks

phansys:
## Subject


phansys:
Because the `null` value is not documented in these docblocks.

greg0ire:
Oh right. We should migrate to `?Reponse $response = null` BTW.

greg0ire:
Thanks @phansys !

PR #616, opened 2019-08-01
Added support for PHPUnit 8

core23:
## Subject

## Changelog

```markdown
### Added
- Added support for PHPUnit 8 for provided `BlockServiceTestCase`
```

### DIFF
--- a/src/Test/BlockServiceTestCase.php
+++ b/src/Test/BlockServiceTestCase.php
@@ -106,7 +106,7 @@ abstract class InternalBlockServiceTestCase extends TestCase
 }
 
 // NEXT_MAJOR: Remove this hack when dropping support for PHPUnit 7
-if (interface_exists('PHPUnit_Framework_MockObject_MockObject')) {
+if (version_compare(\PHPUnit\Runner\Version::id(), '8.0', '<')) {
     abstract class BlockServiceTestCase extends InternalBlockServiceTestCase
     {
         protected function setUp()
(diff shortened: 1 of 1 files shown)


OskarStark:
Why making it internal?

core23:
You should not extend internal classes. This hack is used to provide phpunit with and without return types.

OskarStark:
I know but before it wasn’t internal and I use it to test my custom blocks and this was the initial goal imo. Doing the opposite is a BC break to me, at least in the stable branch.

The hack 👍🏻

src/Test/BlockServiceTestCase.php
```diff
@@ -99,3 +104,22 @@ protected function assertSettings(array $expected, BlockContextInterface $blockC
         $this->assertSame($completeExpectedOptions, $blockSettings);
     }
 }
+
+// NEXT_MAJOR: Remove this hack when dropping support for PHPUnit 7
+if (interface_exists('PHPUnit_Framework_MockObject_MockObject')) {
+    abstract class BlockServiceTestCase extends InternalBlockServiceTestCase
```
core23: We add the original class name here @OskarStark

So this is still BC

core23:
We could also do the same like symfony does: https://github.com/symfony/symfony/pull/32842

src/Test/BlockServiceTestCase.php
```diff
@@ -99,3 +104,22 @@ protected function assertSettings(array $expected, BlockContextInterface $blockC
         $this->assertSame($completeExpectedOptions, $blockSettings);
     }
 }
+
+// NEXT_MAJOR: Remove this hack when dropping support for PHPUnit 7
+if (interface_exists('PHPUnit_Framework_MockObject_MockObject')) {
```
phansys: Why not `version_compare(\PHPUnit\Runner\Version::id(), '8.0', '<')`?

core23: Fixed

PR #617, opened 2019-08-03
Manual merge

greg0ire:


src/Templating/Helper/BlockHelper.php
```diff
@@ -341,7 +341,10 @@ protected function computeAssets(BlockContextInterface $blockContext, array &$st
         ];
     }
 
-    private function startTracing(BlockInterface $block): array
+    /**
+     * @internal since sonata-project/block-bundle 3.16
+     */
+    protected function startTracing(BlockInterface $block): array
```
phansys: The visibility should be private.

src/Templating/Helper/BlockHelper.php
```diff
@@ -341,39 +341,10 @@ protected function computeAssets(BlockContextInterface $blockContext, array &$st
...
-                'age' => 0,
-            ],
-            'assets' => [
-                'js' => [],
-                'css' => [],
-            ],
-        ];
-    }
-
-    private function stopTracing(BlockInterface $block, array $stats): void
```
phansys: All the methods that were marked with `@internal` at 3.x should remain private in 4.x.

greg0ire: Oh ok sorry

src/Templating/Helper/BlockHelper.php
```diff
@@ -341,38 +341,9 @@ protected function computeAssets(BlockContextInterface $blockContext, array &$st
...
-            ],
-            'assets' => [
-                'js' => [],
-                'css' => [],
-            ],
-        ];
-    }
-
+    /**
+     * @internal since sonata-project/block-bundle 3.16
```
phansys: If these `@internal` annotations (which were already resolved using private visibility in 4.x) are intended to be removed in another PR, I think this one is ready.

greg0ire: Yeah that's for the cleanup PR I guess #602

PR #621, opened 2019-08-09
Deprecate "sonata.templating" service

phansys:
## Subject

Deprecate "sonata.templating" service

## Changelog

```markdown
### Deprecated
- "sonata.templating", "sonata.templating.locator" and "sonata.templating.name_parser" services;
- `Sonata\BlockBundle\Templating\TwigEngine` and `Sonata\BlockBundle\Test\FakeTemplating` classes.
```

## To do
    
- [x] Add tests.

### DIFF
--- a/src/Block/Service/AbstractBlockService.php
+++ b/src/Block/Service/AbstractBlockService.php
@@ -19,6 +19,7 @@ use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\OptionsResolver\OptionsResolverInterface;
+use Twig\Environment;
 
 /**
  * @author Sullivan Senechal <soullivaneuh@gmail.com>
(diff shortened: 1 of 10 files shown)


src/Block/Service/AbstractBlockService.php
```diff
@@ -31,33 +32,62 @@ abstract class AbstractBlockService implements BlockServiceInterface
...
+     */
+    private $twig;
+
+    /**
+     * NEXT_MAJOR: Make `$twig` argument mandatory and remove other arguments.
+     *
+     * @param Environment|EngineInterface|string $templatingOrDeprecatedName
      */
-    public function __construct($templatingOrDeprecatedName = null, EngineInterface $templating = null)
+    public function __construct($templatingOrDeprecatedName = null, EngineInterface $templating = null, Environment $twig = null)
```
core23: Why do we need the third parameter?

phansys: Because we want to respect the BC promise for users which are extending this class and need to keep the behavior for `getTemplating()` method. At least, that was my reasoning for adding this argument instead of just using the argument 1.

core23: I prefer using the first parameter only, because we will drop the third parameter with the next major release. So this is just a short living solution.

IMHO we should use the first parameter for this. If the user passes an `Environment` as first parameter, the `getTemplating` method should throw a `RuntimeException`.

phansys: It's fine for me. Could we use `BadMethodCallException` instead of `RuntimeException`?

phansys: Your comments were addressed.

src/Resources/config/block.xml
```diff
@@ -13,33 +13,39 @@
             <tag name="sonata.block"/>
             <argument>sonata.block.container</argument>
             <argument type="service" id="sonata.templating"/>
+            <argument type="service" id="twig"/>
```
core23: You can remove this

phansys: Sure, I forgot those. Done :+1:

src/Test/FakeTemplating.php
```diff
@@ -19,6 +19,8 @@
 /**
  * Mocking class for template usage.
  *
+ * NEXT_MAJOR: Remove this class.
```
core23: You should also deprecate this class

src/Block/Service/AbstractBlockService.php
```diff
@@ -70,7 +98,18 @@ public function __construct($templatingOrDeprecatedName = null, EngineInterface
      */
     public function renderResponse($view, array $parameters = [], Response $response = null)
     {
-        return $this->getTemplating()->renderResponse($view, $parameters, $response);
+        if (!$this->twig) {
```
core23:
```suggestion
        if (null !== $this->twig) {
```

phansys: This check requires exactly the opposite.
`$this->twig` will be lossely evaluated to `false` if an instance of `Environment` is not set. There is already a check at `__construct()` that prevents this property to hold other value than `Environment`.

src/Test/FakeTemplating.php
phansys: It's done.

src/Block/Service/AbstractBlockService.php
core23: Oh sorry, I meant `null === $this->twig`.

I prefer explicit type checks as long as we don't have typed properties

phansys: The explicit type check is emulated with [this check](https://github.com/sonata-project/SonataBlockBundle/pull/621/files#diff-039b4f28b4329f773f4b0fbfda05ea05R80). Regardless that, I've updated this condition as you have proposed.

PR #623, opened 2019-08-19
Fix phpstan findings

core23:
## Subject

## Changelog

```markdown
### Changed
- Removed unused BlockServiceManager constructor arguments

### Fixed
- Added explicit type check 
- Added check before calling deprecated method
```

### DIFF
--- a/src/Block/Service/AbstractBlockService.php
+++ b/src/Block/Service/AbstractBlockService.php
@@ -134,6 +134,12 @@ abstract class AbstractBlockService implements BlockServiceInterface
 
     public function setDefaultSettings(OptionsResolverInterface $resolver)
     {
+        if (!$resolver instanceof OptionsResolver) {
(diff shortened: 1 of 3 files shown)


src/Block/Service/AbstractBlockService.php
```diff
@@ -134,6 +134,12 @@ public function renderPrivateResponse($view, array $parameters = [], Response $r
 
     public function setDefaultSettings(OptionsResolverInterface $resolver)
     {
+        if (!$resolver instanceof OptionsResolver) {
```
OskarStark: Really? I mean didn’t we keep it like this to allow the deprecated one? 🤔

core23: ` configureSettings` needs an `OptionsResolver` instance, but we never checked this

PR #633, opened 2019-09-11
Replace deprecated EngineInterface

core23:
## Subject

## Changelog

```markdown
### Changed
- Replaced deprecated EngineInterface
```


core23:
> Do we need to add a dependency?

We have the dependency: https://github.com/sonata-project/SonataBlockBundle/blob/3.x/composer.json#L40

PR #631, opened 2019-09-05
Merge branch '3.x' into master

phansys:

PR #628, opened 2019-08-23
Deprecate Sonata Cache in favor of PSR Cache

core23:
## Subject

I am targeting this branch, because this is BC.

Closes #627 

## Changelog

```markdown
### Added
- Added support for PSR Cache

### Deprecated
- Deprecate Sonata Cache in favor of PSR Cache
```

PR #635, opened 2019-09-22
DO NOT MERGE

core23:
This is just a test to resolve the conflicts of #631

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,7 @@
     "require": {
         "php": "^7.3",
         "doctrine/common": "^2.3",
+        "psr/cache": "^1.0",
         "sonata-project/cache": "^1.0 || ^2.0",
         "sonata-project/doctrine-extensions": "^1.1",
         "sonata-project/form-extensions": "^1.0",
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
         "php": "^7.3",
         "doctrine/common": "^2.3",
         "psr/cache": "^1.0",
+        "psr/cache-implementation": "*",
         "sonata-project/cache": "^1.0 || ^2.0",
         "sonata-project/doctrine-extensions": "^1.1",
         "sonata-project/form-extensions": "^1.0",


core23:
Adding `psr/cache-implementation` causes the problem. Maybe we could remove this dependency? @phansys

phansys:
`psr/cache-implementation` causes the problem because it is provided by `symfony/cache: 4.3.4`, which has a requirement against [`php: ^7.1.3`]( https://packagist.org/packages/symfony/cache#v4.3.4), while `sonata-project/block-bundle` at `master` branch has the requirement `php: ^7.3`.
BTW, I didn't analyzed yet if requirement for `psr/cache-implementation` can be dropped.

core23:
> `psr/cache-implementation` [...], which has a requirement against [`php: ^7.1.3`](https://packagist.org/packages/symfony/cache#v4.3.4), while `sonata-project/block-bundle` at `master` branch has the requirement `php: ^7.3`.

I don't get the problem. We require `^7.3`, while symfony is requiring `^7.1.3`. Travis runs on PHP 7.3, so this should match both requirements 🤔

phansys:
> > `psr/cache-implementation` [...], which has a requirement against [`php: ^7.1.3`](https://packagist.org/packages/symfony/cache#v4.3.4), while `sonata-project/block-bundle` at `master` branch has the requirement `php: ^7.3`.
> 
> I don't get the problem. We require `^7.3`, while symfony is requiring `^7.1.3`. Travis runs on PHP 7.3, so this should match both requirements

You're right @core23, I misread the requirement as `~7.1.3`, my apologies.
I think we can analyze if `psr/cache-implementation` could be removed, but I also think that we should have a clear idea why we need to do that.

core23:
> I think we can analyze if `psr/cache-implementation` could be removed, but I also think that we should have a clear idea why we need to do that.

The dependency was added with #628. The original idea was to force the user to use a cache, which is wrong. The bundle can work without a cache. 

The non-psr cache was optional: https://github.com/sonata-project/SonataBlockBundle/blob/3.17.0/src/Templating/Helper/BlockHelper.php#L89

phansys:
I agree with the proposed solution, just don't like to not understand why the dependency conflict is in place.

PR #611, opened 2019-07-24
Added phpstan checks

core23:
## Subject

Improve code quality by running phpstan 🚀 

I am targeting this branch, because this is BC, but should only be done for the next major version.

PR #646, opened 2019-11-08
Final major preparations

core23:
## Subject

I am targeting this branch, because this is (hopefully) the last PR to get a new stable release.

Closes #638

## Changelog

```markdown
### Added
- Add `BlockLoaderInterface::exists` method to interface
- Add `BlockContextManagerInterface::exists` method to interface

### Changed
- Make all services public by default

### Removed
- Remove service class parameters
- Remove PageBundle specific code
- Remove addClassesToCompile

### Fixed
- Fix twig import of macros
- Load menu block only when KnpMenuBundle exist
- Remove undefined method call
```

PR #648, opened 2019-11-08
Added phpstan checks

core23:
## Subject

Improve code quality by running phpstan 🚀 

I am targeting this branch, because this is BC, but should only be done for the next major version.

This is the same PR like #611. It looks like GitHub can't add new hooks with existing PRs.

### DIFF
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -1,0 +1,12 @@
+on: [push, pull_request]
+name: Quality assurance
+jobs:
+  phpstan:
+    name: PHPStan
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: PHPStan
+        uses: "docker://oskarstark/phpstan-ga"
+        with:
+          args: analyse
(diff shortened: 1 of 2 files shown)


core23:
It is working @OskarStark  🚀

.github/workflows/qa.yml
```diff
@@ -0,0 +1,12 @@
...
+jobs:
+  phpstan:
+    name: PHPStan
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: PHPStan
+        uses: "docker://oskarstark/phpstan-ga"
+        with:
+            args: analyse
```
FlintCIBot: - yamllint: wrong indentation: expected 10 but found 12 (indentation)

core23:
Depends on #646 which fixes all phpstan errors

phpstan.neon.dist
```diff
@@ -0,0 +1,19 @@
+parameters:
+    level: 2
+
+    paths:
+        - src
+# Uncomment this, when phpstan extensions are allowed
+#        - tests
```
OskarStark: I install all needed extensions it isn’t that working?

core23: The problem is, that we don't require phpunit as a dev-dependency. PHPStan can't find the `TestCase` class

PR #654, opened 2019-11-15
Fix phpstan findings

core23:
## Subject

You can test the release using this repo: https://github.com/core23/sonata-block-major

I am targeting this branch, because this is (hopefully) the last PR for before the major release.

## Changelog

```markdown
### Added
- Add support for twig 3

### Changed
- Make command fully lazy

### Fixed
- Fix some phpstan findings
```

PR #655, opened 2019-11-17
Fix phpstorm fidings

franmomu:
## Subject

Driven by https://github.com/sonata-project/SonataBlockBundle/pull/654, I tried to see what PhpStorm could find before the major release and these are the results. But there are a couple of things:

The `FilterResponseEvent` will be removed in Symfony 5.0. What should I do? Update the `Upgrade` file?
And there is also another interface deprecated in [BlockManagerInterface](https://github.com/sonata-project/SonataBlockBundle/blob/master/src/Model/BlockManagerInterface.php#L19), `PageableManagerInterface`:

```
PageableManagerInterface class is deprecated since 1.3 in favor of 
 'Sonata\DatagridBundle\Pager\PageableInterface
```
But this bundle does't use datagrid-bundle.

I am targeting this branch, because there are BC-break changes.

## Changelog

```markdown
### Changed
- Replaced deprecated FilterResponseEvent
### Fixed
- Fix some PhpStorm findings
```

### DIFF
--- a/src/Block/BlockContextManager.php
+++ b/src/Block/BlockContextManager.php
@@ -128,7 +128,7 @@ final class BlockContextManager implements BlockContextManagerInterface
             'extra_cache_keys' => [],
             'attr' => [],
             'template' => false,
-            'ttl' => (int) $block->getTtl(),
+            'ttl' => $block->getTtl(),
         ]);
 
         $optionsResolver
--- a/src/Block/BlockContextManager.php
+++ b/src/Block/BlockContextManager.php
@@ -141,8 +141,8 @@ final class BlockContextManager implements BlockContextManagerInterface
 
         // add type and class settings for block
         $class = ClassUtils::getClass($block);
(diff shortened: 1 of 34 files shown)


src/Block/BlockContextManager.php
```diff
@@ -141,8 +141,8 @@ private function configureSettings(OptionsResolver $optionsResolver, BlockInterf
 
         // add type and class settings for block
         $class = ClassUtils::getClass($block);
-        $settingsByType = isset($this->settingsByType[$block->getType()]) ? $this->settingsByType[$block->getType()] : [];
```
phansys: These kind of changes should be covered by their respective rule at `.php_cs.dist`. "ternary_to_null_coalescing" in this case.

core23:
Nice job @franmomu 👍

src/Block/BlockContextManager.php
franmomu: Agree, but if we add it, does this mean that flint is going to fail in every PR (that the repository has this kind of code)?

PR #664, opened 2019-12-03
Add missing PHPDoc for shared test class

core23:
## Subject

I am targeting this branch, because this is a pedantic change.


greg0ire:
There are commit that should only exit on master in this PR :thinking:

core23:
> There are commit that should only exit on master in this PR 🤔

Fixed

PR #671, opened 2019-12-27
Remove test deprecation

core23:
## Subject

Just calling the `setUp` method was throwing a deprecation warning. The deprecation is now thrown, when someone accesses the `templating` property.

I am targeting this branch, because we must maintain the old 3.x branch because we still use it for most of our bundles.

## Changelog

```markdown
### Removed
- Removed deprecation warnings when extending `BlockServiceTestCase`
```


greg0ire:
What should be done about Travis?

core23:
> What should be done about Travis?

No idea. Looks like travis is fetching wrong version information. There is no change inside the `composer.json`.

Issue #700, opened 2020-04-24
Upgrade block-bundle to 4.x in sonata-project bundles


wbloszyk:
## Feature Request
After watching some issue with Blocks (https://github.com/sonata-project/SonataMediaBundle/pull/1737, https://github.com/sonata-project/SonataMediaBundle/pull/1732) I came up with an idea.

All sonata-project stable bundles(Sonata 3) using `block-bundle 3.x` but 4.x is released. 

### block-bundle 4.0 upgrade change:

1. block constructor 
1. `AbstractAdminBlockService/AdminBlockServiceInterface` to `EditableBlockService`
1. Return type hints
1. Conflict with CoreBundle

### Resolve something like: (I will add example son - PR for SonataNewsBundle)

1. https://github.com/sonata-project/SonataClassificationBundle/pull/488
1. some steps needed 
    - remove  `AbstractAdminBlockService/AdminBlockServiceInterface` implementations
    - add methods from `AbstractAdminBlockService`
    - add methods from `EditableBlockServce`
    - methods from `EditableBlockServie` should not call `Abstra...` methods and vice versa
    - some check for override methods should be add
1. I will describe it below
1. Resolve conflict with CoreBundle 
    - https://github.com/sonata-project/SonataCoreBundle/issues/739
    - new realease of BlockBundle 4.x will be required (to allow use 0.x extensions)
    - or user can install extensions 0.x as 1.x

### Return type hints return type hints
1. move Block directory to Model/Block
1. Create the same Class in Block like in example: 
```php 
if (`SonataBlockBundle3`) {
    class SomeBlockService Extends Modal/Block/SomeBlockServie {}  
} elseif (`SonataBlockBundle4`) {
    // extends the came thing like Modal/Block/SomeBlockServie
    // we can add `EditableBlockService` becouse user make block-bundle MAJOR UPDATE
    class SomeBlockService Extends AbstractBlockService implements EditableBlockService
    {
        /**
        * var Modal/Block/SomeBlockServie
        */
        private $blockService;

        public function __construct($env, $someService)
        {
              $this->blockService = new Modal/Block/SomeBlockServie($env, $someService),
        }

        // add return type hints 
        public function configureSettings(OptionsResolver $resolver): void
        {
              $this->blockService->configureSettings($resolver);
        }
        .... // for all extended methods
    }  
}
```
## Before start - resolve conflict with CoreBundle: 
[...]

PR #701, opened 2020-05-05
Drop unnesessary constructor

wbloszyk:
## Subject

This change will allow prepare blocks which extends `AbstractAdminBlockService` to use one parameter constructor required in BlockBundle 4.x

I am targeting this branch, because it is PATCH.

### DIFF
--- a/src/Block/Service/AbstractAdminBlockService.php
+++ b/src/Block/Service/AbstractAdminBlockService.php
@@ -17,7 +17,6 @@ use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\BlockBundle\Meta\Metadata;
 use Sonata\BlockBundle\Model\BlockInterface;
 use Sonata\CoreBundle\Validator\ErrorElement;
-use Symfony\Component\Templating\EngineInterface;
 
 @trigger_error(
     'The '.__NAMESPACE__.'\AbstractAdminBlockService class is deprecated since sonata-project/block-bundle 3.16 '.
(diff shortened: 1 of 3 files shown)


src/Block/Service/AbstractAdminBlockService.php
```diff
@@ -32,14 +31,6 @@
  */
 abstract class AbstractAdminBlockService extends AbstractBlockService implements AdminBlockServiceInterface
 {
-    /**
-     * @param string $name
-     */
-    public function __construct($name, EngineInterface $templating)
```
phansys, line 38: The only change I see at first glance is that parent definition is allowing `null` for the 2nd argument.

wbloszyk: abstract class AbstractAdminBlockService extends AbstractBlockService implements AdminBlockServiceInterface
```php
abstract class AbstractBlockService implements BlockServiceInterface
{
    ...

    /**
     * NEXT_MAJOR: Make `$twig` argument mandatory and remove other arguments.
     *
     * @param Environment|EngineInterface|string $templatingOrDeprecatedName
     */
    public function __construct($templatingOrDeprecatedName = null, EngineInterface $templating = null)
    ...
```

phansys: Yes, that's exactly what I'm pointing (`EngineInterface $templating = null`). Although I don't know if it worths.

wbloszyk: This change is require to https://github.com/sonata-project/SonataBlockBundle/issues/700

greg0ire:
Thanks @wbloszyk !

wbloszyk:
@greg0ire Can you release it? I need it in https://github.com/sonata-project/SonataMediaBundle/pull/1740

greg0ire:
Sure!

wbloszyk:
> Sure!

@greg0ire Can you edit `AbstractAdminBlockService` and fix constructor too?
```php
    public function __construct($name, ?EngineInterface $templating)
```

`?` is missing too :(

greg0ire:
There is no constructor in `AbstractAdminBlockService`

wbloszyk:
Oh, this PR fix it. I dont have it in media-bundle becouse it is not release yet. :1st_place_medal: for me. All it is OK.

PR #705, opened 2020-05-22
Fix autowired service id mismatch

core23:
## Subject

This is a backport of a bugfix (#646). The registered block id must match with the service id.

## Changelog

```markdown
### Fixed
- Fix autowired service id mismatch
```


greg0ire:
Please amend your commit message so that it explains how the autowiring is broken, and why your changes fix that situation.

PR #710, opened 2020-06-02
Support doctrine/common 3 in 3.x

jaikdean:
## Subject

This PR adds support for `doctrine/common` 3. It follows the upgrade instructions in Doctrine's [release notes](https://github.com/doctrine/common/releases/tag/3.0.0).

I am targeting this branch, as requested in #709

Also see corresponding PR for https://github.com/sonata-project/SonataAdminBundle/pull/6127.

## Changelog

```markdown
### Added
- Added support for `doctrine/common` 3.
```

PR #725, opened 2020-08-11
4.3.0

phansys:


jordisala1991:
I think we need to check 2 things:

1. 3.x is merged on 4.x
2. add the missing changelog from bugfixes / features from 3.x

This release is a bit harder than a normal one

jordisala1991:
I think the point 1. is done already

jordisala1991:
And for the second point I think we should only mention the support for doctrine/common 3

phansys:
>I think the point 1. is done already

Yes, currently there is nothing to compare [between 4.x and 3.x](https://github.com/sonata-project/SonataBlockBundle/compare/4.x...3.x).

phansys:
>And for the second point I think we should only mention the support for doctrine/common 3

I've updated the changelog through `release` command, with some minimal changes to target the `4.x` branch. Is this mention omitted in the PR where the change was introduced?

jordisala1991:
I dont think so but it was for a Pr targeting 3.x branch and the command is not able to understand that a release for 4.x might contain changelog from Prs on 3.x branch

phansys:
I just add the [changelog comment](https://github.com/sonata-project/SonataBlockBundle/pull/725/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR6-R7) from #710. Please, let me know if there is something else to improve.

jordisala1991:
I think it is great, thank you :)

PR #764, opened 2020-10-09
Allow replace `AbstractAdminBlockService` by `EditableBlockService`

wbloszyk:
## Subject

Based on https://github.com/sonata-project/dev-kit/issues/785 `AbstractAdminBlockService` should be replace by  `extends AbstractBlockService implements EditableBlockService`. To do it this methods should be replaced:
- `buildEditForm` => `configureEditForm`
- `buildCreateForm` => `confiureCreateForm`
- `validateBlock` => `validate`
- `getBlockMetadata` => `getMetadata`

This change will allow move code to new methods with keep BC. In second step in `4.x` branch we can add `AbstractAdminBlockService implement EditableBlockService`. In third step we can add support for BlockBundle v4 in sonata 3 (with some little BC-break).

I am targeting this branch, because this change respect BC.

## Changelog

```markdown
### Added
- Added `Sonata\BlockBundle\Form\Mapper\BlockFormMapper` class for allow use `FormMapper` with AdminBundle
- Added some methods to `Sonata\BlockBundle\Block\Service\AbstractAdminBlockService` for feature compatibility:
    - `configureEditForm()`
    - `configureCreateForm()`
    - `validate()`
    - `getMetadata()`
```

### DIFF
--- a/src/Block/Service/AbstractAdminBlockService.php
+++ b/src/Block/Service/AbstractAdminBlockService.php
@@ -13,8 +13,11 @@ declare(strict_types=1);
 
 namespace Sonata\BlockBundle\Block\Service;
 
-use Sonata\AdminBundle\Form\FormMapper;
+use Sonata\AdminBundle\Form\FormMapper as AdminFormMapper;
+use Sonata\BlockBundle\Form\Mapper\BlockFormMapper;
+use Sonata\BlockBundle\Form\Mapper\FormMapper;
 use Sonata\BlockBundle\Meta\Metadata;
+use Sonata\BlockBundle\Meta\MetadataInterface;
 use Sonata\BlockBundle\Model\BlockInterface;
 use Sonata\Form\Validator\ErrorElement;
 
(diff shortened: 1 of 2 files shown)


VincentLanglet:
> In third step we can add support for BlockBundle v4 in sonata 3 (with some little BC-break).

We should not.
- Every BC-break should be forbidden on current branch
- We should focus on Sonata 4 instead

We just need to support BlockBundle 4 on Sonata 4.

src/Block/Service/AbstractAdminBlockService.php
```diff
@@ -31,45 +34,97 @@
  */
 abstract class AbstractAdminBlockService extends AbstractBlockService implements AdminBlockServiceInterface
```
core23, line 35: This class is already deprecated, no need to change anything right here

core23 (changes requested):

src/Form/Mapper/BlockFormMapper.php
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\BlockBundle\Form\Mapper;
+
+use Sonata\AdminBundle\Form\FormMapper as AdminFormMapper;
```
core23, line 16: We should not need any `AdminBundle` reference. We decoupled this bundle from the admin bundle on the 4.x branch

core23 (changes requested):

src/Form/Mapper/BlockFormMapper.php
wbloszyk: This code is use only when user generate form for Block in Admin site.

VincentLanglet: It would still need to require sonataAdmin on master https://github.com/sonata-project/SonataBlockBundle/blob/master/composer.json

The right way should be to add this code on sonataAdmin instead, if necessary.

core23: AFAIK we only need this mapping for the PageBundle:
https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Mapper/PageFormMapper.php

VincentLanglet: So it could be directly added to the PageBundle then

core23: > So it could be directly added to the PageBundle then

Already done ;)

PR #767, opened 2020-10-11
DevKit updates for 3.x branch

SonataCI:


jordisala1991:
Need help here @OskarStark , why does the phpstan does not pass? The action looks a lot like the psalm, but it is failing on composer install. Maybe it is because it does not have git installed on the docker container? Thats the only difference I could see: https://github.com/psalm/psalm-github-actions/blob/master/Dockerfile#L16

PR #773, opened 2020-10-25
Try build

jordisala1991:


jordisala1991:
This reveals the true problem with php 8: tracked here: #1207

PR #776, opened 2020-10-28
Fixed invalid types

fastnloud:
## Fixed invalid types
Fixed tests and allow `null` as a return value in `getTemplate` (a `string` value is expected here while this may strictly not be the case).

I am targeting this branch, because this return type was added here.

## Changelog

```markdown
### Changed
- `\Sonata\BlockBundle\Block\BlockContextInterface::getTemplate` signature has been changed to allow `null` values
- `\Sonata\BlockBundle\Exception\Strategy\StrategyManager::getBlockRenderer` may now throw an `InvalidArgumentException` if the service does not match the expected interface
- `\Sonata\BlockBundle\Exception\Strategy\StrategyManager::getBlockFilter` may now throw an `InvalidArgumentException` if the service does not match the expected interface

### Deprecated
- Passing a boolean to option "template" is deprecated and will not be allowed in 5.0, pass a string or null instead at `Sonata\BlockBundle\Block\BlockContext`
```


fastnloud:
@core23 @VincentLanglet @wbloszyk is there any chance this can be reviewed? If anything else needs to be done please let me know. I'll be happy to help out.

composer.json
```diff
@@ -52,6 +52,8 @@
         "matthiasnoback/symfony-dependency-injection-test": "^4.1",
         "phpstan/phpstan": "^0.12.29",
         "phpstan/phpstan-symfony": "^0.12.10",
+        "phpunit/phpunit": "^9.4",
+        "psalm/plugin-phpunit": "^0.13.0",
```
VincentLanglet, line 56: Can you explain why you need to add this / update the psalm baseline ?

src/Block/BlockContext.php
```diff
@@ -65,6 +65,6 @@ public function setSetting(string $name, $value): BlockContextInterface
 
     public function getTemplate(): string
     {
-        return $this->getSetting('template');
+        return (string) $this->getSetting('template');
```
VincentLanglet, line 68: What are you trying to fix ?

A `null` template ?

If the setting has something else than `string|null`, this seems to be an error.

src/Exception/Strategy/StrategyManager.php
```diff
@@ -171,7 +171,7 @@ public function getBlockFilter(BlockInterface $block): FilterInterface
      *
      * @throws \RuntimeException
      */
-    private function getFilterService(string $name): object
+    private function getFilterService(string $name): ?object
```
core23, line 174: I would keep the `object` return and throw an exception if the filter is null / undefined

src/Exception/Strategy/StrategyManager.php
```diff
@@ -185,7 +185,7 @@ private function getFilterService(string $name): object
      *
      * @throws \RuntimeException
      */
-    private function getRendererService(string $name): object
+    private function getRendererService(string $name): ?object
```
core23, line 188: Same here

core23 (changes requested):

core23:
Sorry for the delay. The PR looks good, except these two parts

src/Model/BaseBlock.php
```diff
@@ -80,7 +80,7 @@ public function __toString()
         return sprintf('%s ~ #%s', $this->getName(), $this->getId());
     }
 
-    public function setName(string $name): void
+    public function setName(?string $name): void
```
VincentLanglet, line 83: Why this change ?

VincentLanglet (commented):
Some of these change should be done in 3.x

For instance: https://github.com/sonata-project/SonataBlockBundle/blob/3.x/src/Exception/Strategy/StrategyManager.php#L179

src/Exception/Strategy/StrategyManager.php
VincentLanglet: Currently the check is done in `getBlockFilter`.

It could be smarter to move the check inside this function.

src/Block/BlockContext.php
fastnloud: Exactly that. If `null` it'll result in a fatal error because the function expects it to always be a `string` while this may strictly not be the case.

VincentLanglet: It seems better to fix the typehint and update to `?string` then

fastnloud: Well, I noticed that for example the `\Sonata\BlockBundle\Block\Service\AbstractBlockService::execute` requires this to be a `string` so I did not want to break that. Perhaps we can introduce a `hasTemplate` function so this can be checked properly if need be?

src/Model/BaseBlock.php
fastnloud: Because this setter restricts setting a `null` value (name = nullable) and conflicts with the documentation `@var string|null` unless this has been changed.

src/Block/BlockContext.php
VincentLanglet: You mean
```
$this->renderResponse($blockContext->getTemplate(), [
            'block_context' => $blockContext,
            'block' => $blockContext->getBlock(),
        ], $response);
```
?

Both 
```
$this->renderResponse(null, [])
```
and 
```
$this->renderResponse('', [])
```
Will return an error.

So I would say, let's first fix the `getTemplate` return typehint.
And then if some call are not checking `null`, we'll fix them. But these calls wasn't working before so we're not breaking something.

src/Model/BaseBlock.php
VincentLanglet: We have multiple `@var string|null` in Sonata project.
It's generally `null` by default, but then we only can set a value, and never set `null` again.

My question is not clear. I meant "Is there a valid use case for setting `null` here ?". Did you find code setting `null` ?

fastnloud: Thank you for clarifying. A use case would be if the `BaseBlock` model is used for a (custom) block where the `name` field is not required and can be left empty (i.e. is nullable). While this used to be working perfectly fine before this, it breaks now.

Is there any reason to not be able to change the value to its default in this case?

src/Block/BlockContext.php
fastnloud: In that case we can just fix the typehint and update it to `?string` as you suggested.

src/Model/BaseBlock.php
VincentLanglet: > Is there any reason to not be able to change the value to its default in this case?

No, I'm not against this change ; I was just curious.

composer.json
fastnloud: I'll check that again (not really familiar with Psalm). Anyhow, I think we just fixed 2 of those issues by moving the exceptions inside the `getFilterService` and `getRendererService` functions.

VincentLanglet: Yes, now I understand the psalm baseline change.

But now the plugin added since you're not even using it I think.

composer.json
```diff
@@ -52,6 +52,7 @@
         "matthiasnoback/symfony-dependency-injection-test": "^4.1",
         "phpstan/phpstan": "^0.12.29",
         "phpstan/phpstan-symfony": "^0.12.10",
+        "phpunit/phpunit": "^9.4",
```
VincentLanglet, line 55: Should not be needed too.

VincentLanglet (changes requested):

psalm-baseline.xml
```diff
@@ -1,18 +1,2 @@
...
-            <code>$this-&gt;container-&gt;get($this-&gt;renderers[$name])</code>
-        </NullableReturnStatement>
-    </file>
-    <file src="src/Test/BlockServiceTestCase.php">
-        <UndefinedClass occurrences="1">
-            <code>TestCase</code>
-        </UndefinedClass>
-    </file>
-</files>
+<files psalm-version="4.1.1@16bfbd9224698bd738c665f33039fade2a1a3977"/>
```
VincentLanglet, line 2: I think you can remove the psalm baseline if it's empty

composer.json
fastnloud: Then I'd need to add the undefined class line in the `psalm-baseline.xml` file, also `SYMFONY_PHPUNIT_VERSION` is set to `9.4` in `phpunit.xml.dist` and the tests will fail below `9.3.1` due to incompatibility so it makes more sense to add this as dev dependency as well no? What do you think?

VincentLanglet: Then maybe it's better to re-add this to the psalm-baseline (?)

Because it hides a bigger issue.
We're exposing a class in `src/Test` which requires `phpunit/phpunit` without explicitely requiring it.

It should theorically be in the require section, not require-dev then.

But @sonata-project/contributors can see this differently.

fastnloud: I see what you mean. Shouldn't the `BlockServiceTestCase` be moved to the` \Tests` namespace then as it's only used in tests? 🤔

VincentLanglet: > I see what you mean. Shouldn't the `BlockServiceTestCase` be moved to the` \Tests` namespace then as it's only used in tests? 🤔

I think it's not like this because the `test` directory is not exported when you install this bundle ; because of https://github.com/sonata-project/SonataBlockBundle/blob/3.x/.gitattributes#L7

fastnloud: What about adding `@psalm-suppress UndefinedClass` to the `BlockServiceTestCase` so it's (more) clear this class should be ignored? 

All the other errors are fixed so no `psalm-baseline.xml` should then be required.

VincentLanglet: Could be a good temporary solution.

fastnloud: The only usage I can find is in tests so I'm still a bit confused why we should not move it into the `\Tests` namespace but I'm probably missing something.

Running `make test` seems to be working with this change.

VincentLanglet: It's used by others bundle: https://github.com/search?q=org%3Asonata-project+BlockServiceTestCase&type=code

VincentLanglet (dismissed):

src/Block/BlockContext.php
fastnloud: @VincentLanglet I just noticed that the allowed types for `template` are mixed (`['string', 'bool']`) so setting the return type to `?string` doesn't always work.

VincentLanglet: Can you provide me some links about this `bool` value ?

Because when I looked at the usage, 
```
$this->renderResponse($blockContext->getTemplate(), [
            'block_context' => $blockContext,
            'block' => $blockContext->getBlock(),
        ], $response);
```
A bool value would lead to an error...

fastnloud: https://github.com/sonata-project/SonataBlockBundle/blob/9724f2f658b329a69889fcd61d0ba363926d87e4/src/Block/BlockContextManager.php#L176

VincentLanglet: Maybe we should use `template` => null instead then.

fastnloud: It's been there for a while https://github.com/sonata-project/SonataBlockBundle/commit/64b3a61cfd830f4853e369f77b57083c0be20b28 but yeah, setting this to `null` makes more sense just by looking at the interface and the annotations.

VincentLanglet (commented):
Should we add an UPGRADE NOTE?

src/Block/BlockContextManager.php
```diff
@@ -139,7 +139,7 @@ private function configureSettings(OptionsResolver $optionsResolver, BlockInterf
                 ->addAllowedTypes('extra_cache_keys', 'array')
                 ->addAllowedTypes('attr', 'array')
                 ->addAllowedTypes('ttl', 'int')
-                ->addAllowedTypes('template', ['string', 'bool'])
+                ->addAllowedTypes('template', ['null', 'string'])
```
franmomu, line 142: I think this would be BC break for someone having declared the `template` to `true`.

fastnloud: You're right. What do you suggest? We can of course still use `(string) $this->getSetting('template');` in the `getTemplate` method.

VincentLanglet: Since the signature was `getTemplate(): string`, I have some trouble to understand how setting the template to false or true could be valid previously.

VincentLanglet: > You're right. What do you suggest? We can of course still use `(string) $this->getSetting('template');` in the `getTemplate` method.

Imho, we don't have to write `(string) $this->getSetting('template');`  in order to add support for boolean since we want to remove it. This was previously not working with boolean.

The only things we could eventually do is
```
// NEXT_MAJOR: Remove bool.
->addAllowedTypes('template', ['null', 'string', 'bool'])
```
But I still wonder if someone can use false/true without any error.

fastnloud: Not when using the `Sonata\BlockBundle\Block\Service\AbstractBlockService::renderResponse` since this accepts a `string` only and will also crash when `null` is passed to that function.

VincentLanglet (approved):

src/Block/BlockContextManager.php
```diff
@@ -139,7 +139,8 @@ private function configureSettings(OptionsResolver $optionsResolver, BlockInterf
                 ->addAllowedTypes('extra_cache_keys', 'array')
                 ->addAllowedTypes('attr', 'array')
                 ->addAllowedTypes('ttl', 'int')
-                ->addAllowedTypes('template', ['string', 'bool'])
+                // NEXT_MAJOR: Remove bool.
```
phansys, line 142: IMO, we should [deprecate](https://symfony.com/doc/current/components/options_resolver.html#deprecating-the-option) the types we want to remove in the next major.

src/Exception/Strategy/StrategyManager.php
```diff
@@ -155,42 +148,48 @@ public function getBlockRenderer(BlockInterface $block): RendererInterface
...
     {
         if (!isset($this->filters[$name])) {
             throw new \RuntimeException('The filter "%s" does not exist.');
         }
 
-        return $this->container->get($this->filters[$name]);
+        $service = $this->container->get($this->filters[$name]);
+
+        if (!$service instanceof FilterInterface) {
+            throw new \RuntimeException(sprintf('The service "%s" is not an exception filter.', $name));
```
phansys, line 170:
```suggestion
            throw new \InvalidArgumentException(sprintf('The service "%s" is not an exception filter.', $name));
```

src/Exception/Strategy/StrategyManager.php
```diff
@@ -155,42 +148,48 @@ public function getBlockRenderer(BlockInterface $block): RendererInterface
...
     {
         if (!isset($this->renderers[$name])) {
             throw new \RuntimeException('The renderer "%s" does not exist.');
         }
 
-        return $this->container->get($this->renderers[$name]);
+        $service = $this->container->get($this->renderers[$name]);
+
+        if (!$service instanceof RendererInterface) {
+            throw new \RuntimeException(sprintf('The service "%s" is not an exception renderer.', $name));
```
phansys, line 190:
```suggestion
            throw new \InvalidArgumentException(sprintf('The service "%s" is not an exception renderer.', $name));
```

phansys (changes requested):

VincentLanglet (dismissed):

VincentLanglet (dismissed):

src/Block/BlockContextManager.php
```diff
@@ -139,7 +140,16 @@ private function configureSettings(OptionsResolver $optionsResolver, BlockInterf
...
+                ->addAllowedTypes('template', ['null', 'string', 'bool'])
+                // NEXT_MAJOR: Remove setDeprecated.
+                ->setDeprecated('template', 'sonata-project/block-bundle', '4.5.0', static function (Options $options, $value) {
```
phansys, line 146:
```suggestion
                ->setDeprecated('template', 'sonata-project/block-bundle', '4.5.0', static function (Options $options, $value): string {
```

VincentLanglet (dismissed):

fastnloud:
Is the upgrade note something I need to add in this pull requests @phansys since you labeled it?

src/Model/BlockInterface.php
```diff
@@ -35,7 +35,7 @@ public function getId();
     /**
      * Sets the name.
      */
-    public function setName(string $name): void;
+    public function setName(?string $name): void;
```
phansys, line 38: This change breaks BC.

phansys:
> Is the upgrade note something I need to add in this pull requests @phansys since you labeled it?

Yes, you should add an upgrade note at [`UPGRADE-4.x,md`](https://github.com/sonata-project/SonataBlockBundle/blob/b69267260674144fa877935007a12b976328594a/UPGRADE-4.x.md).
Regarding the deprecated option types, I think something like this may be enough.
```markdown
UPGRADE FROM 4.x to 4.x
=======================

### Sonata\BlockBundle\Block\BlockContext

Passing boolean values to "template" option at `BlockContext` is deprecated and
will be not allowed in version 5.0. It will only accept the types `string` and `null`.
```
Please, note that also the **Changelog** section in the PR description must be updated, since you have *Added* new allowed types and *Deprecated* some others.

src/Model/BlockInterface.php
fastnloud: Strictly speaking yes but is this not a bug that was created whilst adding types? I see no reason why this may not be set back to `null` (its original value).

UPGRADE-4.x.md
```diff
@@ -1,2 +1,6 @@
-UPGRADE 4.x
-===========
```
VincentLanglet, line 2: Please, let this title

```
UPGRADE 4.x
===========

UPGRADE FROM 4.4 to 4.5
======================

### Sonata\BlockBundle\Block\BlockContext

Passing a boolean to option "template" is deprecated and will not be allowed in 5.0, pass a `string` or `null` instead.
```

src/Model/BlockInterface.php
phansys: I didn't analyze the reason, but any class implementing this interface will start failing after this change. That's is what I'm pointing here. I think this is not a matter about the strictness level.

fastnloud: @phansys I've reverted this change. It's also a bit unclear to me what the `name` of a block represents or what its intended purpose is. Perhaps someone can clarify? E.g. we've used it as a form field (like `content`) hence we're having trouble updating this to `null`. I can also move that to `settings` instead so the values can be nullified if required. Just by looking at the model itself tho it feels strange that it can be constructed without a name but cannot be reset to `null` via its setter.

Note that the `getTemplate` signature change is still there. That breaks BC too, no? So not sure what we're going to do with that. 

Hopefully we can merge the other fixes, if not, we can still work around it but thought I'd share this first.

phansys: >Note that the getTemplate signature change is still there. That breaks BC too, no?

While we should consider the change at `getTemplate()` as a BC break, at least it will not produce a PHP error.

VincentLanglet:
Seems like this PR was forgotten, can you do a rebase @fastnloud ?

phansys (approved):

VincentLanglet (approved):

VincentLanglet:
Please review @core23 :)

core23 (approved):

VincentLanglet:
Thanks @fastnloud

PR #790, opened 2020-11-27
Add support for PHP 8

core23:
## Changelog

```markdown
### Added
- Added support for PHP 8
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         }
     ],
     "require": {
-        "php": "^7.3",
+        "php": "^7.3 || ^8.0",
         "doctrine/collections": "^1.6",
         "doctrine/common": "^2.7.1 || ^3.0",
         "psr/cache": "^1.0",
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -28,6 +28,7 @@ jobs:
                 php-version:
                     - '7.3'
                     - '7.4'
+                    - '8.0'
                 dependencies: [highest]
                 allowed-to-fail: [false]
                 variant: [normal]
@@ -35,10 +36,6 @@ jobs:
                     - php-version: '7.3'
                       dependencies: lowest
                       allowed-to-fail: false
-                      variant: normal
-                    - php-version: '8.0'
-                      dependencies: highest
-                      allowed-to-fail: true
                       variant: normal
                     - php-version: '7.4'
                       dependencies: highest
@@ -62,10 +59,6 @@ jobs:
 
             - name: Add PHPUnit matcher
               run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
-
-            - name: Configuration required for PHP 8.0
-              if: matrix.php-version == '8.0'
-              run: composer config platform.php 7.4.99
 
             - name: Install variant
               if: matrix.variant != 'normal'


greg0ire:
It's blocked by 2 dependencies.

OskarStark:
Should be done against dev-kit if it works

VincentLanglet:
We need first to support php 8 on doctrine-extensions and form-extensions.

We also need https://github.com/KnpLabs/KnpMenuBundle/issues/437

core23:
This is only possible for the 4.x branch. The 3.x has a conflict with the admin bundle which does not support PHP 8

composer.json
```diff
@@ -20,7 +20,7 @@
         }
     ],
     "require": {
-        "php": "^7.3",
+        "php": "^7.3 || ^8.0",
```
VincentLanglet, line 23:
```suggestion
        "php": ">=7.3",
```

VincentLanglet (changes requested):

composer.json
core23: I can do this, but what's the idea behind this? Can you guarantee that this project will work with PHP 9, 10, 23?

OskarStark: Packagist now recommends using `>=` for libs because we can also not safely say it is not working. So this is more or less blocking users from upgrading.

franmomu: I also prefer having an upper bound, where can we read about this? just curious, the only thing I've found is https://getcomposer.org/doc/faqs/why-are-unbound-version-constraints-a-bad-idea.md

OskarStark: Puh not sure where we had this dicussion in Symfony, and this should only be used for `php` and only for libraries, not for projects, but *not* for required libs ofc.

VincentLanglet: We're already using `>=` for the repository which allowed PHP8, we need to be consistent.

Symfony is using `>=` https://github.com/symfony/symfony/blob/5.x/composer.json#L19.

See https://github.com/symfony/symfony/pull/36876

franmomu: > See [symfony/symfony#36876](https://github.com/symfony/symfony/pull/36876)

Hmmm most of the opinions seem like against the change.

VincentLanglet: > > See [symfony/symfony#36876](https://github.com/symfony/symfony/pull/36876)
> 
> Hmmm most of the opinions seem like against the change.

Phpunit is using it too : https://github.com/sebastianbergmann/phpunit/blob/master/composer.json#L25

Most of people were asking for the support of PHP8 for PHPunit8 which were not supported anymore, and it leads to
https://github.com/sebastianbergmann/phpunit/commit/73435f4bac40a63d18f04707858c738a64c1f7ed ; nothing else was needed. People are even now asking for the PHP8 support for PHPUnit 7.

Also, Php is not following semver https://twitter.com/Ocramius/status/778675003990614017
So `^7.3` could already lead to errors.

greg0ire: I wouldn't use `>=`: following the controversy you linked to , it's now possible to do `--ignore-platform-req=php`. I don't think we need to worry about people not being able to play with our packages on recent versions of PHP anymore.

core23: > [sebastianbergmann/phpunit@73435f4](https://github.com/sebastianbergmann/phpunit/commit/73435f4bac40a63d18f04707858c738a64c1f7ed) ; nothing else was needed. People are even now asking for the PHP8 support for PHPUnit 7.

I wonder why the allow any future PHP version. The guys from the phpcc are very strict in most cases.

Adding strict rules that can be overridden via composer looks good, but allowing ANY version that can cause problems sounds like a bad idea. 

> Symfony is using `>=` https://github.com/symfony/symfony/blob/5.x/composer.json#L19.

Symfony is very lose in some decisions, e.g. phpstan / psalm annotations are not wanted.

Should we start a thread about this on the dev-kit or should I amend my commit?

greg0ire: > We already have >= for [6 repositories]

We are 6 people. I'm taking care of form-extensions.

VincentLanglet: > Should we start a thread about this on the dev-kit or should I amend my commit?

I think we don't need the thread, the decision seems almost taken.

Against `>=` there is already @franmomu, @greg0ire, you.

Maybe @phansys, @jordisala1991 or @wbloszyk could answer here.

But like I said https://github.com/sonata-project/SonataBlockBundle/pull/790#discussion_r554049504 ; there is 6 packages to fix then.

greg0ire: https://github.com/sonata-project/form-extensions/pull/194/commits
https://github.com/sonata-project/cache/pull/174

VincentLanglet (approved):
https://github.com/sonata-project/dev-kit/pull/1350 will be needed

composer.json
VincentLanglet: All 6 PR are now merged.

Does it worth a patch release ?

greg0ire: It is not worth a release at all IMO: if some packages have already been released with `>=`, it's already too late unless we maybe re-tag them (not a huge fan of that).

PR #845, opened 2021-03-07
Fixed invalid types

core23:
## Subject

Backport of #776

## Changelog

```markdown
### Changed
- Extend possible return type of `BlockContextInterface::getTemplate`
```


src/Block/BlockContextInterface.php
```diff
@@ -43,7 +43,7 @@ public function getSetting($name);
     public function setSetting($name, $value);
 
     /**
-     * @return string
+     * @return string|false|null
```
VincentLanglet, line 46: In 4.x it's restricted to string|null.
If we add `false`, I think it's better to add `NEXT_MAJOR: Remove false` to help to solve the conflict.

But I don't know if we should add `false`, or just keep `string|null`.
The implementation is wrong, not the interface.

core23: Yes, the implementation is wrong, but adding just using `string|null` is wrong too for the 3.x branch.

There's no need for a NEXT_MAJOR hint, because it's already solved.

VincentLanglet: > There's no need for a NEXT_MAJOR hint, because it's already solved.

It still will produce conflict when merging 3.x into master.
And the one who gonna solve the conflict could believe that the `false` part should be added.
That's why I was talking about 
> to add `NEXT_MAJOR: Remove false` to help to solve the conflict.

I'm doing a lot of merge and solving a lot of conflict, it's sometimes hard when I not the one who did the PR.

wbloszyk (approved):

PR #861, opened 2021-03-28
Fix phpdoc

VincentLanglet:
## Subject

`&` should be used instead of `|` or we'll end with psalm errors.
(I got these on SonataAdminBundle).

## Changelog

```markdown
### Fixed
- `BlockServiceTestCase` phpdoc
```

### DIFF
--- a/src/Test/BlockServiceTestCase.php
+++ b/src/Test/BlockServiceTestCase.php
@@ -37,7 +37,7 @@ use Twig\Environment;
 abstract class BlockServiceTestCase extends TestCase
 {
     /**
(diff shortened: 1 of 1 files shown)


phansys (approved):

phansys:
Since this change only affects the tests, I'd say it should be considered pedantic.

VincentLanglet:
> Since this change only affects the tests, I'd say it should be considered pedantic.

It's not the `tests` dir but the `src/Test` dir. This class is exposed by the bundle and used by other bundles.
A pedantic means that a release is not needed. Here I need one.

PR #891, opened 2021-07-03
Adapt to interface name

greg0ire:
## Subject

Because of named parameters, the signature of an implementation should
match the one of the interface. I have chose to always use the name in
the interface because that should be a smaller BC-break.

Side-effect: fixing the Psalm job of the CI.

## Changelog

```markdown
### Fixed
- Fix inconsistencies in parameter names between interfaces and their implementations
```

### DIFF
--- a/src/Model/BaseBlock.php
+++ b/src/Model/BaseBlock.php
@@ -160,11 +160,11 @@ abstract class BaseBlock implements BlockInterface
         return $this->updatedAt;
     }
 
-    public function addChildren(BlockInterface $child)
+    public function addChildren(BlockInterface $children)
     {
-        $this->children[] = $child;
+        $this->children[] = $children;
 
-        $child->setParent($this);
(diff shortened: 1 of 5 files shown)


VincentLanglet (approved):

src/Model/BaseBlock.php
```diff
@@ -160,11 +160,11 @@ public function getUpdatedAt()
         return $this->updatedAt;
     }
 
-    public function addChildren(BlockInterface $child)
+    public function addChildren(BlockInterface $children)
```
VincentLanglet, line 163: We should rename the method to `addChild` in next major.

greg0ire: Yes 👍 the implementation makes very little sense with this naming.

jordisala1991 (approved):

VincentLanglet:
Thanks

PR #958, opened 2021-10-08
Deprecate BlockManagerInterface

core23:
## Subject

I am targeting this branch, because this interface is not used anywhere.

## Changelog

```markdown
### Deprecated
- Deprecate `Sonata\BlockBundle\Model\BlockManagerInterface`
```

### DIFF
--- a/src/Model/BlockManagerInterface.php
+++ b/src/Model/BlockManagerInterface.php
@@ -17,6 +17,8 @@ use Sonata\Doctrine\Model\ManagerInterface;
 use Sonata\Doctrine\Model\PageableManagerInterface;
 
 /**
+ * @deprecated since sonata-project/block-bundle 4.x, to be removed in 5.0.
+ *
  * @phpstan-template T of object
  * @phpstan-extends ManagerInterface<T>
  */


composer.json
```diff
@@ -53,6 +52,7 @@
         "phpstan/phpstan": "^0.12.29",
         "phpstan/phpstan-symfony": "^0.12.10",
         "psalm/plugin-symfony": "^2.0",
+        "sonata-project/doctrine-extensions": "^1.10.1",
```
VincentLanglet: Can we remove totally the dependency ? If not, will we be able to do it in next major ?

core23: We still need this for phpstan / psalm, because it is imported in the deprecated class.

VincentLanglet: Ok, so we're be able to remove the dependancy in the next major ; should we add a reminder ?

VincentLanglet (dismissed):

composer.json
core23: I'll do this once the dev-kit job merges this to the 5.x branch

jordisala1991:
https://github.com/search?q=org%3Asonata-project+BlockManagerInterface&type=code

This interface is used on SonataPageBundle

core23:
> This interface is used on SonataPageBundle

But there is no usage and implementation in *this* project. We should move the interface to the `SonataPageBundle`.

VincentLanglet (approved):

VincentLanglet (changes requested):
After more thought I'm not sure about removing the doctrine-extensions from the require section.
Someone using the BlockManagerInterface will have the project crashing if he didn't required the doctrine-extensions package. And he has no reason to do it since it doesn't use it directly...

I feel like it safer to keep it in the require section for this major

core23:
> After more thought I'm not sure about removing the doctrine-extensions from the require section.
> Someone using the BlockManagerInterface will have the project crashing if he didn't required the doctrine-extensions package. And he has no reason to do it since it doesn't use it direct

That's a problem for all optional dependencies we have....

IMHO there are only two case for someone using the `BlockManagerInterface`:
- they should either use the `SonataPageBundle` (which requires the optional dependency) 
- there is a custom project implementation for the interface. This project should then require the `sonata-project/doctrine-extensions` inside the `composer.json` file

VincentLanglet:
> > After more thought I'm not sure about removing the doctrine-extensions from the require section.
> > Someone using the BlockManagerInterface will have the project crashing if he didn't required the doctrine-extensions package. And he has no reason to do it since it doesn't use it direct
> 
> That's a problem for all optional dependencies we have....

The issue here is that `doctrine-extensions` wasn't considered as an optional dependency before.
Changing his state from `required` to `optional` is the BC-break I point out.
I don't think it's an issue to keep it in the require section until we remove the dependency... is it ?

> IMHO there are only two case for someone using the `BlockManagerInterface`:
> 
> * there is a custom project implementation for the interface. This project should then require the `sonata-project/doctrine-extensions` inside the `composer.json` file

Dunno what others @sonata-project/contributors thinks about this BC-break.
But if we accepts this, it should be at least in the changelog/upgrade section.

jordisala1991:
> > > After more thought I'm not sure about removing the doctrine-extensions from the require section.
> 
> > > Someone using the BlockManagerInterface will have the project crashing if he didn't required the doctrine-extensions package. And he has no reason to do it since it doesn't use it direct
> 
> > 
> 
> > That's a problem for all optional dependencies we have....
> 
> 
> 
> The issue here is that `doctrine-extensions` wasn't considered as an optional dependency before.
> 
> Changing his state from `required` to `optional` is the BC-break I point out.
> 
> I don't think it's an issue to keep it in the require section until we remove the dependency... is it ?
> 
> 
> 
> > IMHO there are only two case for someone using the `BlockManagerInterface`:
> 
> > 
> 
> > * there is a custom project implementation for the interface. This project should then require the `sonata-project/doctrine-extensions` inside the `composer.json` file
> 
> 
> 
> Dunno what others @sonata-project/contributors thinks about this BC-break.
> 
> But if we accepts this, it should be at least in the changelog/upgrade section.
> 
> 

My main concern would be about providing the required migration path for other bundles using this class before declaring it deprecated.

core23:
> My main concern would be about providing the required migration path for other bundles using this class before declaring it deprecated.

There is no migration path, because this bundle does not provide any implementation for this method.

This [PR](https://github.com/sonata-project/SonataPageBundle/pull/1369) adds the interface to the only bundle that is implementing the interface.

composer.json
```diff
@@ -25,7 +25,6 @@
         "doctrine/common": "^2.7.1 || ^3.0",
         "psr/cache": "^1.0 || ^2.0 || ^3.0",
         "sonata-project/cache": "^2.0",
-        "sonata-project/doctrine-extensions": "^1.10.1",
```
VincentLanglet: Please review @sonata-project/contributors 

Does moving a require to a require-dev is a BC-break ?

core23: Ping @sonata-project/contributors

dmaicher: I don't think we should cover this under our BC promises. If someone is relying on this package then they need to require it explicitly, or?

VincentLanglet: But here you may need it without relying on this package directly.
If you were using Sonata\BlockBundle\Model\BlockManagerInterface, you'll need the dependency.

core23: Move this to a new PR: https://github.com/sonata-project/SonataBlockBundle/pull/971

VincentLanglet (approved):

PR #971, opened 2021-11-01
Make sonata-project/doctrine-extensions an optional dependency

core23:
## Subject

Supports https://github.com/sonata-project/SonataBlockBundle/pull/958

The only class that requires the dependency is deprecated.

## Changelog

```markdown
### Changed
- `sonata-project/doctrine-extensions` is an optional dependency now
```


VincentLanglet (dismissed):

phansys (commented):
Even if the class requiring this package is deprecated, I think we should add an entry in the `suggest` section about this dependency.

core23:
> Even if the class requiring this package is deprecated, I think we should add an entry in the `suggest` section about this dependency.

Not a big fan of this. We should not advertise the usage of a deprecated functionality.

We only use the `suggests` section to recommend other feature-rich projects (e.g. menu, admin, messenger) not for such small atomic methods.

phansys:
>Not a big fan of this. We should not advertise the usage of a deprecated functionality.

I'm not happy suggesting a dependency for a deprecated feature neither, but I think the situation is tricky and can be considered a bug by users trying to use a feature that requires a dependency which is not installed nor suggested.
BTW, don't take my concern as a blocker.

VincentLanglet:
> > Not a big fan of this. We should not advertise the usage of a deprecated functionality.
> 
> I'm not happy suggesting a dependency for a deprecated feature neither, but I think the situation is tricky and can be considered a bug by users trying to use a feature that requires a dependency which is not installed nor suggested. BTW, don't take my concern as a blocker.

IMHO I'm not a big fan of removing a dependency when we deprecate something.
The user could use the deprecated class without any occurence of the removed dependency and ends up with a crash.
Maybe instead of a suggest, we should at least add some line in the upgrade note.

core23:
> Maybe instead of a suggest, we should at least add some line in the upgrade note.

Done

VincentLanglet (dismissed):

UPGRADE-4.x.md
```diff
@@ -1,6 +1,13 @@
 UPGRADE 4.x
 ===========
 
+UPGRADE FROM 4.7 to 4.x
+=======================
+
+### `sonata-project/doctrine-extensions` is optional
+
+By deprecating `Sonata\BlockBundle\Model\BlockManagerInterface`, the doctrine extension library is now optional.
```
phansys:
```suggestion
By deprecating `Sonata\BlockBundle\Model\BlockManagerInterface`, the doctrine extensions library is now optional.
```

phansys (dismissed):

VincentLanglet (approved):

phansys (approved):

phansys:
Thank you @core23!

PR #998, opened 2021-12-19
POC - Use Symfony cache

### LINKED ISSUE #663: Remove Sonata Cache
We should use symfony cache instead.

Even the doctrine cache was just deprecated in favor of symfony cache:
https://github.com/doctrine/DoctrineCacheBundle#deprecation-warning

VincentLanglet:
## Subject

Closes #663.

## Changelog

```markdown
### Added
- Support for Symfony Cache

### Deprecated
- Support for Sonata Cache
```


src/DependencyInjection/SonataBlockExtension.php
```diff
@@ -69,6 +69,11 @@ public function load(array $configs, ContainerBuilder $container): void
         $loader->load('exception.xml');
         $loader->load('commands.xml');
 
+        // NEXT_MAJOR: Remove this check and the legacy_core.xml file.
+        if (class_exists(\Sonata\Cache\CacheManagerInterface::class)) {
```
VincentLanglet, line 73: I added this in order to keep the old behavior by overriding the service if the user installed SonataCache.

src/Resources/config/core.xml
```diff
@@ -31,7 +31,7 @@
             <argument type="service" id="sonata.block.renderer"/>
             <argument type="service" id="sonata.block.context_manager"/>
             <argument type="service" id="event_dispatcher"/>
-            <argument type="service" id="sonata.cache.manager" on-invalid="ignore"/>
+            <argument type="service" id="TODO" on-invalid="ignore"/>
```
VincentLanglet, line 34: I dunno what to use as id for the symfony cache, help @sonata-project/contributors

src/Templating/Helper/BlockHelper.php
```diff
@@ -114,23 +117,26 @@ class BlockHelper
...
+        $cacheItemPool = null,
         ?HttpCacheHandlerInterface $cacheHandler = null,
         ?Stopwatch $stopwatch = null
     ) {
         $this->blockServiceManager = $blockServiceManager;
         $this->cacheBlocks = $cacheBlocks;
         $this->blockRenderer = $blockRenderer;
         $this->eventDispatcher = $eventDispatcher;
-        $this->cacheManager = $cacheManager;
+        $this->cacheItemPool = $cacheItemPool;
```
VincentLanglet, line 139: I need to add a deprecation here.

src/Resources/config/core.xml
wbloszyk: sonata.symfony_cache

VincentLanglet: I don't think that's right, sonata shouldn't be in the id I'm looking for.
Event dispatcher from symfony has id `event_dispatcher`.
Here, I'm looking for the symfony service implementing the `CacheItemPoolInterface` interface.

VincentLanglet: I also need to replace the `sonata.cache.noop` occurences.
https://github.com/sonata-project/SonataBlockBundle/search?q=sonata.cache.noop

wbloszyk: My bad. In this place we should inject service by alias. Then alias can be configure in files core/legacy_core. Also deprecation can be easy add in legacy_core. In 5.x corrrct service can be inject directly. Then `sonata.cache_manager_alias` can be used

VincentLanglet:
I think I made a mistake.
`CacheManagerInterface` cannot be replaced by CacheItemPoolInterface, but more by something like container, which return cache services.

src/Resources/config/core.xml
VincentLanglet: I think I made a huge mistake thinking CacheManagerInterface could be replaced by CacheItemPoolInterface.
But a CacheManagerInterface returns CacheServices, and CacheServices can be replaced by CacheItemPoolInterface.

So the cache_manager_alias, should be something like the container in the symfony case. This way with a service id, it returns the cache service with this id. But still I don't know what's the "recorder" in this case and how to deal with extra_cache_keys...

PR #1011, opened 2022-01-29
Deprecate caching related functionality

### LINKED ISSUE #663: Remove Sonata Cache

dmaicher:
## Deprecate all caching related functionality 

I am targeting this branch, because its BC.

Closes https://github.com/sonata-project/SonataBlockBundle/issues/663

## Changelog

```markdown
### Deprecated
- integration with `SonataCacheBundle` has been deprecated
- interface `Sonata\BlockBundle\Cache\HttpCacheHandlerInterface` and it's implementations have been deprecated
- constructor Signature of `Sonata\BlockBundle\Templating\Helper\BlockHelper` changed: all caching related arguments should be removed
- all caching configuration options have been deprecated and `sonata_block.http_cache` should be set to `false`
```

## To do

- [ ] Update docs?
- [x] Update `BlockHelper` constructor signature
- [x] Add changelog

### DIFF
--- a/src/DependencyInjection/Compiler/BlockHelperCompilerPass.php
+++ b/src/DependencyInjection/Compiler/BlockHelperCompilerPass.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\BlockBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Parameter;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * @internal
+ *
+ * NEXT_MAJOR: remove this class
+ */
+final class BlockHelperCompilerPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        $cacheBlocks = $container->getParameter('sonata_block.cache_blocks');
+        if (0 === \count($cacheBlocks)) {
+            return;
+        }
+
+        @trigger_error(
+            'Defining cache blocks is deprecated since sonata-project/block-bundle 4.x and will not be supported anymore in 5.0.',
+            \E_USER_DEPRECATED
+        );
+
+        $blockHelperDefinition = $container->getDefinition('sonata.block.templating.helper');
+        $blockHelperDefinition->setArguments([
+            new Reference('sonata.block.manager'),
+            new Parameter('sonata_block.cache_blocks'),
+            new Reference('sonata.block.renderer'),
+            new Reference('sonata.block.context_manager'),
+            new Reference('event_dispatcher'),
+            new Reference('sonata.cache.manager', ContainerInterface::NULL_ON_INVALID_REFERENCE),
+            new Reference('sonata.block.cache.handler', ContainerInterface::NULL_ON_INVALID_REFERENCE),
+            new Reference('debug.stopwatch', ContainerInterface::NULL_ON_INVALID_REFERENCE),
+        ]);
+    }
+}
--- a/src/DependencyInjection/Compiler/BlockHelperCompilerPass.php
+++ b/src/DependencyInjection/Compiler/BlockHelperCompilerPass.php
@@ -34,7 +34,7 @@ final class BlockHelperCompilerPass implements CompilerPassInterface
         $hasCacheBlocks = false;
         foreach ($cacheBlocks as $blocks) {
             foreach ($blocks as $cacheType) {
-                $hasCacheBlocks = $hasCacheBlocks || $cacheType !== 'sonata.cache.noop';
+                $hasCacheBlocks = $hasCacheBlocks || 'sonata.cache.noop' !== $cacheType;
             }
         }
 
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\BlockBundle\DependencyInjection;
 
+use Symfony\Component\Config\Definition\BaseNode;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
@@ -28,6 +29,13 @@ final class Configuration implements ConfigurationInterface
      * @var array<string, string>
      */
     private $defaultContainerTemplates;
+
+    /**
+     * NEXT_MAJOR: remove this member.
+     *
+     * @var bool
+     */
+    private $httpCacheDisabled = false;
 
     /**
      * @param array<string, string> $defaultContainerTemplates
@@ -49,11 +57,21 @@ final class Configuration implements ConfigurationInterface
             ->fixXmlConfig('block')
             ->fixXmlConfig('block_by_class')
             ->validate()
-                ->always(static function ($value) {
-                    foreach ($value['blocks'] as $name => &$block) {
+                ->always(function (&$value) {
+                    foreach ($value['blocks'] as &$block) {
                         if (0 === \count($block['contexts'])) {
                             $block['contexts'] = $value['default_contexts'];
                         }
+                    }
+
+                    // NEXT_MAJOR: remove this block
+                    if (true !== $this->httpCacheDisabled) {
+                        @trigger_error(
+                            'Not setting the "sonata_block.http_cache" config option to false is deprecated since sonata-project/block-bundle 4.x and will fail in 5.0.',
+                            \E_USER_DEPRECATED
+                        );
+                    } else {
+                        $value['http_cache'] = false;
                     }
 
                     return $value;
@@ -73,10 +91,24 @@ final class Configuration implements ConfigurationInterface
                 ->end()
 
                 ->scalarNode('context_manager')->defaultValue('sonata.block.context_manager.default')->end()
+                // NEXT_MAJOR: deprecate option and only allow setting it to false
                 ->arrayNode('http_cache')
                     ->addDefaultsIfNotSet()
+                    ->beforeNormalization()
+                        ->always(function ($v) {
+                            if (false === $v) {
+                                $this->httpCacheDisabled = true;
+
+                                return [];
+                            }
+
+                            return $v;
+                        })
+                    ->end()
                     ->children()
+                        // NEXT_MAJOR: remove option
                         ->scalarNode('handler')->defaultValue('sonata.block.cache.handler.default')->end()
+                        // NEXT_MAJOR: remove option
                         ->booleanNode('listener')->defaultTrue()->end()
                     ->end()
                 ->end()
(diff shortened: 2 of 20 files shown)


src/DependencyInjection/Configuration.php
```diff
@@ -76,8 +93,14 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->arrayNode('http_cache')
```
VincentLanglet, line 93: Then we can deprecate this node isn't it ?

src/DependencyInjection/Configuration.php
```diff
@@ -76,8 +93,14 @@ public function getConfigTreeBuilder(): TreeBuilder
...
-                        ->scalarNode('handler')->defaultValue('sonata.block.cache.handler.default')->end()
-                        ->booleanNode('listener')->defaultTrue()->end()
+                        // NEXT_MAJOR: deprecate option and only allow it to be set to "sonata.block.cache.handler.noop".
+                        ->scalarNode('handler')
+                            ->defaultValue('sonata.block.cache.handler.default')
+                        ->end()
+                        // NEXT_MAJOR: deprecate option and only allow it to be set to false.
+                        ->booleanNode('listener')
+                            ->defaultTrue()
+                        ->end()
```
VincentLanglet: Shouldn't it be `NEXT_MAJOR: Remove those nodes` ?

src/Templating/Helper/BlockHelper.php
```diff
@@ -134,6 +138,18 @@ public function __construct(
         $this->blockContextManager = $blockContextManager;
         $this->cacheHandler = $cacheHandler;
         $this->stopwatch = $stopwatch;
+
+        // NEXT_MAJOR: remove this check
+        if ($cacheManager !== null) {
```
VincentLanglet: We need to update the signature to allow passing the HttpCacheHandlerInterface as 6th param and Stopwatch as 7th.

tests/App/config.yml
```diff
@@ -18,3 +18,8 @@ services:
     Sonata\BlockBundle\Tests\App\Controller\DemoController:
         tags:
             - controller.service_arguments
+
+sonata_block:
+    http_cache:
+        handler: sonata.block.cache.handler.noop
+        listener: false
```
VincentLanglet: It's kinda sad to ask people adding some extra options to avoid deprecation, and then removing the option in next major. I dunno if we have another solution...

Do we have another solution @sonata-project/contributors ?

dmaicher: Currently the http cache is enabled by default. So to have the same behavior as on 5.0 people should somehow disable it I think. This would lead to less surprises when upgrading to 5.0.

Maybe someone has a better idea? We could also just allow disabling it by setting `http_cache: false` maybe. At least thats simpler.

dmaicher: We would deprecate the `http_cache` option then on 5.x and remove on 6.0

src/Resources/config/core.xml
```diff
@@ -32,7 +32,7 @@
             <argument type="service" id="sonata.block.context_manager"/>
             <argument type="service" id="event_dispatcher"/>
             <argument type="service" id="sonata.cache.manager" on-invalid="ignore"/>
-            <argument type="service" id="sonata.block.cache.handler.default" on-invalid="ignore"/>
+            <argument type="service" id="sonata.block.cache.handler" on-invalid="ignore"/>
```
dmaicher, line 35: this was a bug. It never used the configured handler service alias... :see_no_evil:

src/Templating/Helper/BlockHelper.php
dmaicher: actually its even more complicated... also the second argument `$cacheBlocks` will disappear

src/DependencyInjection/SonataBlockExtension.php
```diff
@@ -116,17 +117,19 @@ public function configureContext(ContainerBuilder $container, array $config): vo
...
      *
      * @param array<string, mixed> $config
      */
     public function configureCache(ContainerBuilder $container, array $config): void
     {
-        $container->setAlias('sonata.block.cache.handler', $config['http_cache']['handler']);
+        if (is_array($config['http_cache'])) {
+            $container->setAlias('sonata.block.cache.handler', $config['http_cache']['handler']);
 
-        if (null !== $config['http_cache']['listener']) {
```
dmaicher, line 130: this was also a bug: it was not disabled by setting it to `false` :see_no_evil:

tests/App/config.yml
```diff
@@ -18,3 +18,7 @@ services:
     Sonata\BlockBundle\Tests\App\Controller\DemoController:
         tags:
             - controller.service_arguments
+
+# NEXT_MAJOR: remove this block
+sonata_block:
+    http_cache: false
```
dmaicher, line 24: so my idea for the upgrade path would be to encourage users to upgrade to this config to disabled http caching fully.

In 5.x we can then make this a no-op as it will not do anything anymore and also deprecate it. On 6.0 it can be removed then.

WDYT?

src/Templating/Helper/BlockHelper.php
```diff
@@ -134,6 +138,18 @@ public function __construct(
         $this->blockContextManager = $blockContextManager;
         $this->cacheHandler = $cacheHandler;
         $this->stopwatch = $stopwatch;
+
+        // NEXT_MAJOR: remove this check
+        if (null !== $cacheManager) {
+            @trigger_error(
+                sprintf(
+                    'Passing an instance of "%s" as argument 6 for method "%s" is deprecated since sonata-project/block-bundle 4.x. The argument will be removed in 5.0.',
```
VincentLanglet, line 235: Since the argument will be removed, 
people shouldn't do
```
new BlockHelper($blockServiceManager, $cacheBlocks, $blockRenderer, $blockContextManager, $eventDispatcher, null, $cacheHandler, $stopwatch)
```
but
```
new BlockHelper($blockServiceManager, $cacheBlocks, $blockRenderer, $blockContextManager, $eventDispatcher, $cacheHandler, $stopwatch)
```
so we need to allow this syntax if we want to provide a BC upgrade.

Also, `cacheBlocks` and `cacheHandler` need to be deprecated. So the new signature will be
```
new BlockHelper($blockServiceManager, $blockRenderer, $blockContextManager, $eventDispatcher, $stopwatch)
```

So we need something like
```
public function __construct(
        BlockServiceManagerInterface $blockServiceManager,
        $blockRendererOrCacheBlocks,
        object $blockContextManagerOrBlockRenderer,
        object $eventDispatcherOrBlockContextManager,
        ?object $stopwatchOrEventDispatcher = null,
        ?CacheManagerInterface $cacheManager = null,
        ?HttpCacheHandlerInterface $cacheHandler = null,
        ?Stopwatch $stopwatch = null
    ) {
```
and the check can be done on `blockRendererOrCacheBlocks` if it's an array (deprecated) or a blockRenderer.

tests/App/config.yml
VincentLanglet: I think we don't have a better way, so I'm good with this.

VincentLanglet (changes requested):

src/Templating/Helper/BlockHelper.php
VincentLanglet: But then, for the service definition I wonder how to avoid the deprecation, should we have to service definition and using one or the other based on the config ?

VincentLanglet:
Do you have time to finish this pr @dmaicher ?

I think the biggest point is the signature of `BlockHelper::construct` and the config of this service.

dmaicher:
> Do you have time to finish this pr @dmaicher ?
> 
> I think the biggest point is the signature of `BlockHelper::construct` and the config of this service.

@VincentLanglet I will look into it on the weekend

src/Templating/Helper/BlockHelper.php
dmaicher: added a compiler pass to update to the new signature in case the array of cache blocks is empty

src/DependencyInjection/Compiler/BlockHelperCompilerPass.php
```diff
@@ -0,0 +1,51 @@
...
+            @trigger_error(
+                'Defining cache blocks is deprecated since sonata-project/block-bundle 4.x and will not be supported anymore in 5.0.',
+                \E_USER_DEPRECATED
+            );
+
+            return;
+        }
+
+        $blockHelperDefinition = $container->getDefinition('sonata.block.templating.helper');
+        $blockHelperDefinition->setArguments([
```
VincentLanglet, line 51: This is a nice idea.
I just feel like we should do the opposite.

 - Update our 'default' config to pass the service needed by the new signature
 - In this compiler pass, if there is no cache blocks, return.
 - In this compiler pass, if there are cache blocks, trigger a deprecation and update the definition.

This way, in the next major we just have to remove this (instead having to update the service config too).

VincentLanglet (changes requested):

src/DependencyInjection/Compiler/BlockHelperCompilerPass.php
dmaicher: Indeed that's better 👍

dmaicher: changed it

tests/Templating/Helper/BlockHelperTest.php
```diff
@@ -19,15 +19,38 @@
...
+            $this->createMock(BlockServiceManagerInterface::class),
+            [],
+            $this->createMock(BlockRendererInterface::class),
+            $this->createMock(BlockContextManagerInterface::class),
+            $this->createMock(EventDispatcherInterface::class),
+            $this->createMock(CacheManagerInterface::class),
+            $this->createMock(HttpCacheHandlerInterface::class),
+            new Stopwatch()
+        );
+        $this->addToAssertionCount(1);
```
VincentLanglet: This is internal and reported by psalm.

Instead you can use the phpdoc
```
@doesNotPerformAssertions
```

VincentLanglet (changes requested):

VincentLanglet (dismissed):
Dunno if we should mention in the doc that cache is deprecated
image

VincentLanglet:
Please take a look @jordisala1991 :)

jordisala1991 (dismissed):

dmaicher:
Fixing conflicts :yum:

src/DependencyInjection/Compiler/BlockHelperCompilerPass.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    public function process(ContainerBuilder $container): void
+    {
+        /** @var array{by_class: array<class-string, string>, by_type: array<string, string>} $cacheBlocks */
+        $cacheBlocks = $container->getParameter('sonata_block.cache_blocks');
+
+        $hasCacheBlocks = false;
+        foreach ($cacheBlocks as $blocks) {
+            foreach ($blocks as $cacheType) {
+                $hasCacheBlocks = $hasCacheBlocks || $cacheType !== 'sonata.cache.noop';
```
dmaicher: @VincentLanglet after your phpstan changes I realized simply checking `count($cacheBlocks)` is not enough as the array will indeed always have the two keys.

I now tested it on one of my apps and this check works. By default `sonata.cache.noop` is used which basically disables caching.

VincentLanglet (approved):

VincentLanglet:
Thanks !

mremi:
Hi, why remove this nice feature allowing to have a cache strategy by block? I guess for getting rid SonataCacheBundle but I think we should still support the feature

VincentLanglet:
> Hi, why remove this nice feature allowing to have a cache strategy by block? I guess for getting rid SonataCacheBundle but I think we should still support the feature

Cf https://github.com/sonata-project/SonataBlockBundle/issues/663
- The SonataCacheBundle is deprecated/abandoned
- The current implementation is over complicated

If someone is still interested in this feature, PR are welcomed to add a proper implementation with the SymfonyCache instead.

PR #1012, opened 2022-01-30
Deprecate null template for BlockContext

### LINKED ISSUE #855: Return only string in BlockContextInterface::getTemplate
## Feature Request

Ref: https://github.com/sonata-project/SonataBlockBundle/pull/776

I don't know for sure if `getTemplate` should always return a `string`, this is just a reminder to take a look at this for  `5.x`.

VincentLanglet:
## Subject

I am targeting this branch, because BC.

Closes #855.

## Changelog

```markdown
### Deprecated
- Not passing a string value for the template setting of a BlockContext
```

### DIFF
--- a/src/Model/BaseBlock.php
+++ b/src/Model/BaseBlock.php
@@ -75,6 +75,7 @@ abstract class BaseBlock implements BlockInterface
         $this->children = [];
     }
 
+    #[\ReturnTypeWillChange]
     public function __toString()
     {
         return sprintf('%s ~ #%s', $this->getName(), $this->getId());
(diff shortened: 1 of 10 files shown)


src/Model/BaseBlock.php
```diff
@@ -13,6 +13,7 @@
 
 namespace Sonata\BlockBundle\Model;
 
+
```
jordisala1991: We are not using uses for this. Prefer prefix with \

jordisala1991 (dismissed):

jordisala1991 (dismissed):

VincentLanglet:
I added a NEXT_MAJOR comment for the toString method, and found some missing typehint in final class, so I added them. @jordisala1991

jordisala1991 (approved):

core23 (approved):

PR #1039, opened 2022-04-19
deprecate more caching related things within BlockContextManager

dmaicher:
## Subject

Deprecate `$cacheBlocks` argument on `BlockContextManager::__construct`. 

I overlooked this in https://github.com/sonata-project/SonataBlockBundle/pull/1011.

## Changelog

```markdown
### Deprecated
- deprecated passing array `$cacheBlocks` as 3rd argument for `BlockContextManager::__construct`
- deprecated `BlockContextManagerInterface::CACHE_KEY`

```

### DIFF
--- a/src/Block/BlockContextManager.php
+++ b/src/Block/BlockContextManager.php
@@ -45,9 +45,11 @@ final class BlockContextManager implements BlockContextManagerInterface
     private $settingsByClass = [];
 
     /**
+     * NEXT_MAJOR: remove.
+     *
      * @var array{by_class: array<class-string, string>, by_type: array<string, string>}
      */
-    private $cacheBlocks;
+    private $cacheBlocks = ['by_class' => [], 'by_type' => []];
 
     /**
      * @var LoggerInterface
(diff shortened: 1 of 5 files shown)


src/Resources/config/core.php
```diff
@@ -46,7 +46,6 @@
         ->args([
             new ReferenceConfigurator('sonata.block.loader.chain'),
             new ReferenceConfigurator('sonata.block.manager'),
-            '%sonata_block.cache_blocks%',
             (new ReferenceConfigurator('logger'))->nullOnInvalid(),
```
VincentLanglet, line 49: should we keep the logger as argument ? You'll end with a non solvable deprecation isnt it ?

src/Block/BlockContextManager.php
```diff
@@ -45,28 +45,64 @@ final class BlockContextManager implements BlockContextManagerInterface
...
+        } else {
+            throw new \TypeError('Argument 3 must be null|array|LoggerInterface');
+        }
+
+        // NEXT_MAJOR: remove
+        if (null !== $logger) {
+            $this->logger = $logger;
+            @trigger_error(
+                sprintf(
+                    'Passing an instance of "%s" as argument 4 to method "%s" is deprecated since sonata-project/block-bundle 4.x. The argument will be removed in 5.0.',
```
VincentLanglet, line 99: it will be argument 3 if the `$cacheBlocksOrLogger instanceof LoggerInterface`

src/Block/BlockContextManager.php
```diff
@@ -45,28 +45,64 @@ final class BlockContextManager implements BlockContextManagerInterface
...
+        } elseif ($cacheBlocksOrLogger instanceof LoggerInterface) {
+            $this->logger = $cacheBlocksOrLogger;
+        } elseif (null === $cacheBlocksOrLogger) {
+            $this->logger = new NullLogger();
+        } else {
+            throw new \TypeError('Argument 3 must be null|array|LoggerInterface');
+        }
+
+        // NEXT_MAJOR: remove
+        if (null !== $logger) {
```
VincentLanglet, line 95: It's never null, since you're setting to `new NullLogger();`

tests/Block/BlockContextManagerTest.php
```diff
@@ -122,8 +128,7 @@ public function testWithInvalidSettings(): void
         ]);
         $block->expects(static::once())->method('getSetting')->with('template')->willReturn('custom.html.twig');
 
-        $cacheBlock = ['by_class' => [], 'by_type' => []];
-        $manager = new BlockContextManager($blockLoader, $serviceManager, $cacheBlock, $logger);
+        $manager = new BlockContextManager($blockLoader, $serviceManager, $logger);
```
VincentLanglet, line 131:
```suggestion
        $manager = new BlockContextManager($blockLoader, $serviceManager);
```
?

VincentLanglet (changes requested):

src/Resources/config/core.php
dmaicher: having logger as 3rd argument is the new signature and that's fine

src/Block/BlockContextManager.php
dmaicher: the 4th argument will be removed. We need logger to be passed as 3rd argument

src/Block/BlockContextManager.php
dmaicher: hm where am I doing that? :blush: this can happen if someone passes logger as 4th argument

tests/Block/BlockContextManagerTest.php
dmaicher, line 131: no the 3rd argument as logger is fine and even required in this test. It performs assertions on method calls for the logger.

src/Block/BlockContextManager.php
dmaicher: so `$cacheBlocksOrLogger instanceof LoggerInterface` is the new signature.

src/Block/BlockContextManager.php
VincentLanglet: My bad, I messed up between $this->logger and $logger.

src/Block/BlockContextManager.php
VincentLanglet: Oh, I thought the logger was also removed ; ok

VincentLanglet (approved):

src/Block/BlockContextManager.php
franmomu, line 95: Just a comment, maybe we could check with `func_num_args` in these kind of deprecations to avoid also passing `null`.

franmomu (approved):

src/Block/BlockContextManager.php
dmaicher: yeah will keep it in mind for next time :+1:

PR #1041, opened 2022-04-21
start cleaning up deprecated caching functionality

dmaicher:
Start cleaning up caching related deprecations for 5.x

More PRs will follow.


src/Templating/Helper/BlockHelper.php
```diff
@@ -115,142 +87,16 @@ class BlockHelper
      */
     private $stopwatch;
 
-    /**
-     * NEXT_MAJOR: remove the deprecated signature and cleanup the constructor.
-     *
-     * @param array{by_class: array<class-string, string>, by_type: array<string, string>}|BlockRendererInterface $blockRendererOrCacheBlocks
-     */
     public function __construct(
-        BlockServiceManagerInterface $blockServiceManager,
```
dmaicher: @VincentLanglet I noticed that we also need to remove the first argument :cry:  Otherwise its not used anymore at all...

So I guess we need another PR on 4.x to adjust the deprecation of the constructor signature? :confused:

VincentLanglet: Indeed :'(

dmaicher: will look into it :see_no_evil:

src/Templating/Helper/BlockHelper.php
```diff
@@ -487,12 +200,8 @@ public function render($block, array $options = []): string
             $stats['cache']['lifetime'] = $stats['cache']['age'] + $stats['cache']['ttl'];
```
dmaicher: someone has an idea how useful those stopwatch traces/stats are now without the caching functionalities? 

Do we need the `cache` part of the trace?

```
 * @phpstan-type Trace = array{
 *     name: string,
 *     type: string,
 *     duration: int|float|false,
 *     memory_start: int|false,
 *     memory_end: int|false,
 *     memory_peak: int|false,
 *     cache: array{
 *         keys: mixed[],
 *         contextual_keys: mixed[],
 *         handler: false,
 *         from_cache: false,
 *         ttl: int,
 *         created_at: int|false|null,
 *         lifetime: int,
 *         age: int,
 *     },
 *     assets: array{
 *         js: string[],
 *         css: string[],
 *     }
 * }
 ```

Comment (VincentLanglet) [reply src/Templating/Helper/BlockHelper.php]: Maybe it should be rename 'response' since it's not related to cache anymore.
Dunno how usefull it is...

Comment (dmaicher) [reply src/Templating/Helper/BlockHelper.php]: Don't see anything caching related in the profiler:

Screenshot from 2022-04-22 15-59-24

Will make sure I don't break it :blush:

File: src/Resources/views/Profiler/block.html.twig [hunk 3]
```diff
@@ -112,21 +109,6 @@
                 <td>{{ block.duration|number_format(2) }} ms</td>
             </tr>
 
-            {% if block.cache.handler %}
```
dmaicher, line 115: @VincentLanglet so this is never true without the caching functionality. 

I don't think this is meaningful anymore. So I propose to cleanup the trace data and remove the caching related data.

VincentLanglet: Ok

dmaicher:
This first part is ready for review then. Without caching being used this does not break anything. Tried on one of my apps - works fine.

src/Model/BaseBlock.php
```diff
@@ -202,10 +202,6 @@ public function hasParent(): bool
      */
     public function getTtl(): int
     {
-        if (false === $this->getSetting('use_cache', true)) {
-            return 0;
-        }
-
```
franmomu, line 208: Based on the comment, should this method be completely removed? 🤔

dmaicher: Yes 👍 I will continue cleaning up in a follow up PR. Did not want to create one huge PR. Or would you prefer I do it all in one?

VincentLanglet (approved):

src/Model/BaseBlock.php
franmomu: yeah It's fine, didn't realise that the `ttl` property is not removed either 👍

franmomu (approved):

PR #1042, opened 2022-04-21
more caching related deprecations

dmaicher:
## Subject

Some more deprecations needed on 4.x. See https://github.com/sonata-project/SonataBlockBundle/pull/1041#pullrequestreview-948850433

## Changelog

```markdown
### Changed
- marked `BlockHelper::__construct` as `@internal`
### Deprecated
- deprecated passing an instance of `BlockServiceManagerInterface` as first argument for `BlockHelper::__construct`
- deprecated `BaseBlock::$ttl`
```

### DIFF
--- a/src/Model/BaseBlock.php
+++ b/src/Model/BaseBlock.php
@@ -64,6 +64,10 @@ abstract class BaseBlock implements BlockInterface
     protected $type;
 
     /**
+     * NEXT_MAJOR: remove.
+     *
+     * @deprecated since sonata-project/block-bundle 4.x and will be removed in 5.0
+     *
      * @var int|null
      */
     protected $ttl;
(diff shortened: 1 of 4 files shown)


VincentLanglet (dismissed):

franmomu (dismissed):
I haven't check if there are more usages, but if `BlockHelper` is only used as Twig Runtime, we could maybe mark the constructor as `@internal` since the user is not supposed to instantiate it directly.

dmaicher:
> I haven't check if there are more usages, but if `BlockHelper` is only used as Twig Runtime, we could maybe mark the constructor as `@internal` since the user is not supposed to instantiate it directly.

I had a look and seems its only used for Twig extension and Symfony profiler. At least when using it with SonataAdminBundle. Not sure if other bundles use it as well?

Shall we additionally mark it `@internal` on 4.x then? Or 5.x?

@VincentLanglet wdyt?

VincentLanglet:
I'm ok with internal

PR #1050, opened 2022-04-27
resolve NEXT_MAJOR comments

dmaicher:


src/Block/BlockContextManager.php
```diff
@@ -106,8 +105,12 @@ public function get($meta, array $settings = []): BlockContextInterface
                 $e->getMessage()
             ));
 
-            // NEXT_MAJOR: Only pass the template value if it's a string.
-            $settings = $this->resolve($block, $settings + ['template' => $block->getSetting('template')]);
+            $template = $block->getSetting('template');
+
+            $settings = $this->resolve(
+                $block,
+                $settings + (\is_string($template) ? ['template' => $template] : [])
```
dmaicher, line 112: is this what was meant by `NEXT_MAJOR: Only pass the template value if it's a string.` @VincentLanglet ?

jordisala1991 (approved):

src/Block/BlockContextManager.php
VincentLanglet: Seems like. But I kinda forget...
There was https://github.com/sonata-project/SonataBlockBundle/pull/1012 and then https://github.com/sonata-project/SonataBlockBundle/pull/1026/files.

The goal was to stop returning null from the getTemplate method.
And the setting was coming from here.

So it seems ok like this.

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #1052, opened 2022-04-29
Increase PHPStan and Psalm

jordisala1991:
## Subject

I am targeting this branch, because applying this to 5.x is easier due to deprecated code removed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
         "psalm/plugin-symfony": "^3.0",
         "sonata-project/doctrine-extensions": "^1.10.1",
         "symfony/browser-kit": "^4.4 || ^5.3 || ^6.0",
-        "symfony/debug": "^4.4 || ^5.3 || ^6.0",
+        "symfony/error-handler": "^4.4 || ^5.3 || ^6.0",
         "symfony/phpunit-bridge": "^6.0",
         "symfony/stopwatch": "^4.4 || ^5.3 || ^6.0",
         "vimeo/psalm": "^4.1"
--- a/src/Templating/Helper/BlockHelper.php
+++ b/src/Templating/Helper/BlockHelper.php
@@ -138,6 +138,9 @@ class BlockHelper
     {
         $eventName = sprintf('sonata.block.event.%s', $name);
 
+        /**
+         * @psalm-var BlockEvent
+         */
         $event = $this->eventDispatcher->dispatch(new BlockEvent($options), $eventName);
 
         $content = '';
--- a/src/Templating/Helper/BlockHelper.php
+++ b/src/Templating/Helper/BlockHelper.php
@@ -138,9 +138,6 @@ class BlockHelper
     {
         $eventName = sprintf('sonata.block.event.%s', $name);
 
-        /**
-         * @psalm-var BlockEvent
-         */
         $event = $this->eventDispatcher->dispatch(new BlockEvent($options), $eventName);
 
         $content = '';
@@ -150,6 +147,11 @@ class BlockHelper
         }
 
         if (null !== $this->stopwatch) {
+            /**
+             * @psalm-suppress ArgumentTypeCoercion
+             *
+             * @see https://github.com/psalm/psalm-plugin-symfony/pull/254
+             */
             $this->eventTraces[uniqid('', true)] = [
                 'template_code' => $name,
                 'event_name' => $eventName,
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,5 +1,6 @@
 includes:
     - phpstan-baseline.neon
+    - vendor/phpstan/phpstan/conf/bleedingEdge.neon
 
 parameters:
     level: 8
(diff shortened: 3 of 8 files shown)


jordisala1991:
There are a lot of deprecated code on 4.x that makes difficult to increase psalm there, wdyt on doing it directly for 5.x?

jordisala1991:
Not sure how to deal with:

```
ERROR: DocblockTypeContradiction - src/Block/BlockLoaderChain.php:52:13 - Docblock-defined type array<string, mixed> for $configuration is always array (see https://psalm.dev/155)
        if (!\is_string($configuration) && !\is_array($configuration)) {


ERROR: DocblockTypeContradiction - src/Block/BlockLoaderChain.php:71:13 - Docblock-defined type array<string, mixed> for $configuration is always array (see https://psalm.dev/155)
        if (!\is_string($configuration) && !\is_array($configuration)) {


ERROR: RedundantConditionGivenDocblockType - src/Block/BlockServiceManager.php:89:17 - Docblock-defined type Sonata\BlockBundle\Block\Service\BlockServiceInterface for $service is always object (see https://psalm.dev/156)
                \is_object($service) ? \get_class($service) : \gettype($service)


ERROR: DocblockTypeContradiction - src/Block/BlockServiceManager.php:89:63 - Docblock-defined type Sonata\BlockBundle\Block\Service\BlockServiceInterface for $service is always object (see https://psalm.dev/155)
                \is_object($service) ? \get_class($service) : \gettype($service)


ERROR: DocblockTypeContradiction - src/Block/Loader/ServiceLoader.php:47:13 - Docblock-defined type array<string, mixed> for $configuration is always array (see https://psalm.dev/155)
        if (!\is_string($configuration) && !\is_array($configuration)) {


ERROR: DocblockTypeContradiction - src/Block/Loader/ServiceLoader.php:81:13 - Docblock-defined type array<string, mixed> for $configuration is always array (see https://psalm.dev/155)
        if (!\is_string($configuration) && !\is_array($configuration)) {
```

Psalm is right, PHPDOC says we pass array or string and we are checking for it because its only phpdoc , no real typing. For PHPStan we enable the "Do not trust phpdoc" flag, but what should we do for Psalm?

VincentLanglet:
https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/4.x/psalm.xml#L15-L19

?

src/Templating/Helper/BlockHelper.php
```diff
@@ -138,6 +138,9 @@ public function renderEvent(string $name, array $options = []): string
     {
         $eventName = sprintf('sonata.block.event.%s', $name);
 
+        /**
+         * @psalm-var BlockEvent
```
jordisala1991: Not sure if this is the best way to do it. PHPStan detects correctly the type, but psalm not. So if I do a normal assert it fails phpstan.

VincentLanglet: A PR can be made to add the stub https://github.com/phpstan/phpstan-symfony/blob/1.2.x/stubs/Symfony/Component/EventDispatcher/EventDispatcherInterface.stub
in the psalm-symfony plugin
https://github.com/psalm/psalm-plugin-symfony/tree/master/src/Stubs/common/Component

Do you want to do it or I do ?

dmaicher:
Can you wait for my merge and NEXT_MAJOR cleanup? There might be more stuff to fix :see_no_evil:

jordisala1991:
> Can you wait for my merge and NEXT_MAJOR cleanup? There might be more stuff to fix 🙈

Ofc I can wait, anyway think of it as an improvement to catch errors made during deprecation PRs, so getting an error on a cleanup PR might save time of bug fixing later

src/Templating/Helper/BlockHelper.php
jordisala1991: Can you do it? anyway I think we can have this and maybe remove later? Given it is a problem with a missing Stub, shall we consider this a suppress instead? so when it gets fixed we will get an error to help us removing this code. wdyt?

dmaicher:
Everything is merged now. Can you rebase?

dmaicher (dismissed):

dmaicher (dismissed):

composer.json
```diff
@@ -56,7 +56,7 @@
         "psalm/plugin-symfony": "^3.0",
         "sonata-project/doctrine-extensions": "^1.10.1",
         "symfony/browser-kit": "^4.4 || ^5.3 || ^6.0",
-        "symfony/debug": "^4.4 || ^5.3 || ^6.0",
+        "symfony/error-handler": "^4.4 || ^5.3 || ^6.0",
```
jordisala1991, line 59: Probably there are some other deps that need adjustment: psr/container needs to be added on 4.x, psr/cache needs to be removed on 5.x for example.

I am using:

https://github.com/maglnet/ComposerRequireChecker
https://github.com/composer-unused/composer-unused

both installed globally on my machine to test it (it is not possible to add them to dev kit atm, because they also give false positive or don't take into account optional deps)

src/Templating/Helper/BlockHelper.php
VincentLanglet: https://github.com/psalm/psalm-plugin-symfony/pull/254 ; psalm-suppress might indeed help us to remove this code later

jordisala1991: Done, changed to a suprress instead.

dmaicher (approved):

PR #1060, opened 2022-04-29
Drop older versions

jordisala1991:
## Subject

I am targeting this branch, because this is for the next major version.

## Changelog

```markdown
### Removed
- Remove support for Symfony ^4.4 and 5.3.x
- Remove support for Twig ^2
- Remove support for `doctrine/common` ^2
```


jordisala1991:
This is what I would like to do, but not sure if this is the correct place:

1. Drop old versions of doctrine/common and twig (I think this is the correct place, we have done like this on other majors)
2. Drop older sf versions (this can probably be done on 4.x, but is the correct move or should we wait for 5.0?)

VincentLanglet:
I would drop only on 5.x IMHO.

jordisala1991:
So, you agree on all the changes proposed here?

VincentLanglet:
> So, you agree on all the changes proposed here?

I'm in favor of dropping support for symfony 4.4.

For twig and doctrine/common, I don't know if we win anything by dropping the support.
Also, I dunno if some user could have a real reason to be stuck with doctrine common 2 or twig 2.
So, I would say I'm ok with the drop, but don't know the impact.

jordisala1991:
Dev-kit PR: https://github.com/sonata-project/dev-kit/pull/2014

phpstan.neon.dist
```diff
@@ -14,3 +14,5 @@ parameters:
     checkMissingVarTagTypehint: true
     checkMissingTypehints: true
     checkUninitializedProperties: true
+    scanFiles:
+        - vendor/symfony/dependency-injection/Loader/Configurator/ContainerConfigurator.php
```
VincentLanglet, line 18: Why is this needed ? Vendor is generally always scanned isn't it ?

jordisala1991: https://github.com/phpstan/phpstan-symfony#configuration

If I don't add this I can't use `service`, `param` or `abstract_arg` since phpstan reports them as not existent symbols.

jordisala1991:
Should we also drop php7? (of course in another PR)

VincentLanglet:
> Should we also drop php7? (of course in another PR)

I think we can.

VincentLanglet (approved):

PR #1087, opened 2022-07-19
Allow to use collection for block childrens

VincentLanglet:
## Subject

Closes #{put_issue_number_here}.

## Changelog

```markdown
### Added
- BlockInterface::addChild()
- BlockInterface::hasChild()

### Changed
- BlockInterface::getChildren() now returns iterable

### Deprecated
- BlockInterface::addChildren()
- BlockInterface::hasChildren()
```

### DIFF
--- a/src/Model/BaseBlock.php
+++ b/src/Model/BaseBlock.php
@@ -13,6 +13,8 @@ declare(strict_types=1);
 
 namespace Sonata\BlockBundle\Model;
 
+use Doctrine\Common\Collections\Collection;
+
 /**
  * Base abstract Block class that provides a default implementation of the block interface.
  */
@@ -44,7 +46,9 @@ abstract class BaseBlock implements BlockInterface
     protected $parent;
 
     /**
-     * @var BlockInterface[]
+     * NEXT_MAJOR: Restrict typehint to Collection.
+     *
+     * @var Collection<int, BlockInterface>|array<BlockInterface>
      */
     protected $children;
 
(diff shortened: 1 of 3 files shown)


src/Model/BaseBlock.php
```diff
@@ -44,7 +46,9 @@ abstract class BaseBlock implements BlockInterface
     protected $parent;
 
     /**
-     * @var BlockInterface[]
+     * NEXT_MAJOR: Restrict typehint to Collection.
+     *
+     * @var Collection<int, BlockInterface>|array<BlockInterface>
```
jordisala1991, line 51: what is the advantage over iterable?

src/Model/BaseBlock.php
```diff
@@ -170,14 +174,32 @@ public function getUpdatedAt(): ?\DateTime
...
 
+    public function addChild(BlockInterface $child): void
+    {
+        $this->children[] = $child;
+
+        $child->setParent($this);
+    }
+
+    /**
+     * NEXT_MAJOR: Remove this method.
```
jordisala1991, line 185: do we use addChild on page bundle?

src/Model/BaseBlock.php
VincentLanglet: We're using `count`: https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Model/BaseBlock.php#L224

src/Model/BaseBlock.php
VincentLanglet: No.

In the same way we have `addChildren` method in the PageInterface of PageBundle https://github.com/sonata-project/SonataPageBundle/blob/fed26c06856de8d6b1161d0f23001db7b8377e2b/src/Model/PageInterface.php#L189 which is not used. We could rename this to `addChild`.

If you had in mind to remove the addChild method. I'm not chocked having some add/set method in the interface/class which will only be used by the user.

src/Model/BaseBlock.php
dmaicher: so `array<BlockInterface>` would be fine then? :thinking:

VincentLanglet: The goal is to allow PageBundle to use Collection, https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Entity/BaseBlock.php#L34

See https://github.com/sonata-project/SonataPageBundle/issues/1442 or https://github.com/sonata-project/SonataPageBundle/pull/1469#discussion_r923635402

jordisala1991 (approved):

dmaicher (approved):

PR #1089, opened 2022-07-22
Solve next major comments

VincentLanglet:


jordisala1991 (dismissed):

VincentLanglet:
I need to do some work on RecursiveBlockIterator::hasChildren

src/Util/RecursiveBlockIterator.php
```diff
@@ -48,6 +45,6 @@ public function getChildren(): self
 
     public function hasChildren(): bool
     {
-        return $this->current()->hasChildren();
+        return $this->current()->hasChild();
     }
```
VincentLanglet, line 49: hasChildren is in the RecursiveArrayIterator class.

So I wonder Reverting the deprecation for hasChildren -> hasChild in the BlockInterface to use the same method name. WDYT @jordisala1991 ? Do you prefer hasChild or hasChildren ?

jordisala1991: which one is more correct? I think if it complicates things we have more things to improve before namings.

VincentLanglet: In 4.x we had
```
addChildren()
hasChildren()
```

I deprecated them in favor of 
```
addChild() // Add ONE child
hasChild() // Has at least ONE child (in the implementation)
```

I was wondering if `hasChildren` should have been kept, since
there is a RecursiveBlockIterator::hasChildren method, coming from
the RecursiveArrayIterator.
https://www.php.net/manual/fr/recursivearrayiterator.haschildren.php

But looking at https://www.php.net/manual/fr/recursivearrayiterator.haschildren.php#116569, 
the hasChildren is misleading: it's not "Does the iterator has at least one child", but "Does the
iterator is able to have children (even an empty array is allowed)".

So I would say our new naming might be better.

jordisala1991 (approved):

PR #1090, opened 2022-07-26
Avoid hasChildren deprecation

jordisala1991:
## Subject

## Changelog

```markdown
### Fixed
- Avoid deprecation for the `hasChildren` method.
```

### DIFF
--- a/src/Model/BaseBlock.php
+++ b/src/Model/BaseBlock.php
@@ -199,7 +199,7 @@ abstract class BaseBlock implements BlockInterface
         $children->setParent($this);
     }
 
-    public function getChildren(): iterable
+    public function getChildren()
     {
         return $this->children;
     }
--- a/src/Model/BlockInterface.php
+++ b/src/Model/BlockInterface.php
@@ -13,6 +13,8 @@ declare(strict_types=1);
 
 namespace Sonata\BlockBundle\Model;
 
+use Doctrine\Common\Collections\Collection;
+
 /**
  * @method void addChild(self $child)
  * @method bool hasChild()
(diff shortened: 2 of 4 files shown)


VincentLanglet (dismissed):

VincentLanglet:
Since we're rely on the fact we're iterating on Blocks, I removed the `@phpstan-template T` on 5.x
see https://github.com/sonata-project/SonataBlockBundle/blob/5.x/src/Util/RecursiveBlockIterator.php#L22

jordisala1991:
sad I will need to open the code to fix this... tomorrow then.

VincentLanglet (dismissed):

src/Model/BaseBlock.php
```diff
@@ -199,7 +199,7 @@ public function addChildren(BlockInterface $children): void
         $children->setParent($this);
     }
 
-    public function getChildren(): iterable
+    public function getChildren()
```
jordisala1991, line 202: Had to change this to fix phpstan

src/Model/BlockInterface.php
```diff
@@ -92,9 +94,9 @@ public function addChildren(self $children): void;
     /**
      * NEXT_MAJOR: Restrict typehint to Collection.
      *
-     * @return iterable<BlockInterface> $children
+     * @return Collection<int, BlockInterface>|array<BlockInterface> $children
      */
-    public function getChildren(): iterable;
+    public function getChildren();
```
jordisala1991, line 99: And this.

VincentLanglet: I think you can keep the return type.
```
public function getChildren(): iterable;
```

VincentLanglet (approved):

src/Model/BlockInterface.php
jordisala1991: It breaks phpstan.

PR #1093, opened 2022-08-05
Change typehint to BlockInterface

jordisala1991:
## Subject

I don't know why, but psalm complains if I don't change this line. Here are 2 examples:

```
ERROR: InvalidArgument - src/Entity/Transformer.php:182:30 - Argument 1 of addChild expects static, Sonata\PageBundle\Model\PageBlockInterface provided (see https://psalm.dev/004)
            $block->addChild($this->loadBlock($child, $page));
```

```
ERROR: InvalidArgument - src/Entity/BlockInteractor.php:145:61 - Argument 1 of addChild expects static, Sonata\PageBundle\Model\PageBlockInterface provided (see https://psalm.dev/004)
            $blocks[$block->getParent()->getId()]->addChild($block);
```

With this changes, Psalm is happy (I think we will need to add checks with 5.x branch for PageBundle.

## Changelog

```markdown
### Changed
- Change typehint of `addChild` to `BlockInterface` instead of `self`.
```

### DIFF
--- a/src/Model/BlockInterface.php
+++ b/src/Model/BlockInterface.php
@@ -16,7 +16,7 @@ namespace Sonata\BlockBundle\Model;
 use Doctrine\Common\Collections\Collection;
 
 /**
- * @method void addChild(self $child)
+ * @method void addChild(BlockInterface $child)
  * @method bool hasChild()
  */
 interface BlockInterface


VincentLanglet:
We have 
```
public function addChildren(self $children): void;
```
and I assume we will have the same for `addChild()` in 5.x.

I think we should try the 5.x branch on PageBundle to see if the issue will be here.
If not, it's a psalm issue with `@method`.

jordisala1991:
It works with 5.x. It is probably a problem with psalm @method annotation. Even thought it should be reported to psalm, isn't it better to fix it here and not get blocked?

VincentLanglet:
> It works with 5.x. It is probably a problem with psalm @method annotation. Even thought it should be reported to psalm, isn't it better to fix it here and not get blocked?

Sure, I just wanted to be sure we won't have the issue on 5.x

VincentLanglet (approved):

PR #1098, opened 2022-08-12
Remove editable from ContainerBlockService

jordisala1991:
## Subject

I am targeting this branch, because this is bug fixing.

Helps with: https://github.com/sonata-project/SonataPageBundle/pull/1552#pullrequestreview-1070687221

The enabled field is already here: 
https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Admin/BlockAdmin.php#L160-L164
https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Admin/SharedBlockAdmin.php#L81

So it not needs to be added on each block, This was the only one defining this field.

## Changelog

```markdown
### Removed
- Remove `editable` field from `ContainerBlockService`, it was added twice.
```

### DIFF
--- a/src/Block/Service/ContainerBlockService.php
+++ b/src/Block/Service/ContainerBlockService.php
@@ -41,8 +41,6 @@ final class ContainerBlockService extends AbstractBlockService implements Editab
 
     public function configureEditForm(FormMapper $form, BlockInterface $block): void
     {
-        $form->add('enabled');
-
         $form->add('settings', ImmutableArrayType::class, [
             'keys' => [
                 ['code', TextType::class, [


VincentLanglet (approved):

VincentLanglet:
Some question @jordisala1991 

- Does this block should be moved to PageBundle if it's only used there ?
- What about removing `enabled` in the PageBundle instead ? What's the difference ?

jordisala1991:
> Some question @jordisala1991
> 
> * Does this block should be moved to PageBundle if it's only used there ?
> * What about removing `enabled` in the PageBundle instead ? What's the difference ?

About the first point, I would not move anything for now, because with that logic, all blocks that extend EditableBlockInterface should be moved inside SonataPageBundle, but that might not be a good solution, TBH I don't really know.

The enabled on the PageBundle is placed only on the BlockAdmin, not on each block. If I remove it there, I will need to add on each block the enabled, which is not a good idea.

VincentLanglet:
Ok, let's go this way then

jordisala1991:
Do you have time for a release?

VincentLanglet:
> Do you have time for a release?

Sure

PR #1100, opened 2022-08-16
Fix code needed for SonataPageBundle

jordisala1991:
## Subject

I am targeting this branch, because this is BC? Not sure if adding an extra parameter on the interface is, or not.

## Changelog

```markdown
### Added
- Add `removeChild` to be consistent with `addChild` for Block Model.
```

### DIFF
--- a/src/Block/Service/ContainerBlockService.php
+++ b/src/Block/Service/ContainerBlockService.php
@@ -61,7 +61,11 @@ final class ContainerBlockService extends AbstractBlockService implements Editab
             'translation_domain' => 'SonataBlockBundle',
         ]);
 
-        $form->add('children', CollectionType::class);
+        $form->add('children', CollectionType::class, [], [
+            'edit' => 'inline',
+            'inline' => 'table',
+            'sortable' => 'position',
+        ]);
     }
 
     public function execute(BlockContextInterface $blockContext, ?Response $response = null): Response
--- a/src/Block/Service/ContainerBlockService.php
+++ b/src/Block/Service/ContainerBlockService.php
@@ -61,11 +61,7 @@ final class ContainerBlockService extends AbstractBlockService implements Editab
             'translation_domain' => 'SonataBlockBundle',
         ]);
 
-        $form->add('children', CollectionType::class, [], [
-            'edit' => 'inline',
-            'inline' => 'table',
-            'sortable' => 'position',
-        ]);
+        $form->add('children', CollectionType::class);
     }
 
     public function execute(BlockContextInterface $blockContext, ?Response $response = null): Response
--- a/src/Form/Mapper/FormMapper.php
+++ b/src/Form/Mapper/FormMapper.php
@@ -41,10 +41,11 @@ interface FormMapper
     /**
      * @param class-string<FormTypeInterface>|null $type
      * @param array<string, mixed>                 $options
+     * @param array<string, mixed>                 $fieldDescriptionOptions
      *
      * @return static
      */
-    public function add(string $name, ?string $type = null, array $options = []);
+    public function add(string $name, ?string $type = null, array $options = [], array $fieldDescriptionOptions = []);
 
     /**
      * @return static
--- a/src/Form/Mapper/FormMapper.php
+++ b/src/Form/Mapper/FormMapper.php
@@ -41,11 +41,10 @@ interface FormMapper
     /**
      * @param class-string<FormTypeInterface>|null $type
      * @param array<string, mixed>                 $options
-     * @param array<string, mixed>                 $fieldDescriptionOptions
      *
      * @return static
      */
-    public function add(string $name, ?string $type = null, array $options = [], array $fieldDescriptionOptions = []);
+    public function add(string $name, ?string $type = null, array $options = []);
 
     /**
      * @return static
--- a/src/Model/BaseBlock.php
+++ b/src/Model/BaseBlock.php
@@ -199,6 +199,26 @@ abstract class BaseBlock implements BlockInterface
         $children->setParent($this);
     }
 
+    public function removeChild(BlockInterface $child): void
+    {
+        // NEXT_MAJOR: Remove this condition, children will be a Collection.
+        if (\is_array($this->children)) {
+            $key = array_search($child, $this->children, true);
+
+            if (false === $key) {
+                return;
+            }
+
+            unset($this->children[$key]);
+
+            return;
+        }
+
+        if ($this->children->contains($child)) {
+            $this->children->removeElement($child);
+        }
+    }
+
     public function getChildren()
     {
         return $this->children;
--- a/src/Model/BlockInterface.php
+++ b/src/Model/BlockInterface.php
@@ -17,6 +17,7 @@ use Doctrine\Common\Collections\Collection;
 
 /**
  * @method void addChild(BlockInterface $child)
+ * @method void removeChild(BlockInterface $child)
  * @method bool hasChild()
  */
 interface BlockInterface
--- a/tests/Util/RecursiveBlockIteratorIteratorTest.php
+++ b/tests/Util/RecursiveBlockIteratorIteratorTest.php
@@ -28,6 +28,9 @@ final class RecursiveBlockIteratorIteratorTest extends TestCase
 
         $block1->addChild($block2);
         $block1->addChild($block3);
+        $block1->removeChild($block3);
+        $block1->removeChild($block4);
+        $block1->addChild($block3);
 
         $i = new RecursiveBlockIteratorIterator([$block1, $block4]);
 


src/Block/Service/ContainerBlockService.php
```diff
@@ -61,7 +61,11 @@ public function configureEditForm(FormMapper $form, BlockInterface $block): void
             'translation_domain' => 'SonataBlockBundle',
         ]);
 
-        $form->add('children', CollectionType::class);
+        $form->add('children', CollectionType::class, [], [
```
jordisala1991: This is needed so the collection renders as inline and starts working:
image

src/Form/Mapper/FormMapper.php
```diff
@@ -41,10 +41,11 @@ public function reorder(array $keys);
     /**
      * @param class-string<FormTypeInterface>|null $type
      * @param array<string, mixed>                 $options
+     * @param array<string, mixed>                 $fieldDescriptionOptions
      *
      * @return static
      */
-    public function add(string $name, ?string $type = null, array $options = []);
+    public function add(string $name, ?string $type = null, array $options = [], array $fieldDescriptionOptions = []);
```
jordisala1991: I think we might have a problem here, once I added the 4th parameter on containerBlockService, phpstan complains because that parameter does not exist on the interface, and he is right.

but fieldDescriptionOptions have the followin type on AdminBundle: 

```
 * @psalm-type FieldDescriptionOptions array{
 *     accessor?: string|callable|PropertyPathInterface,
 *     associated_property?: string|callable|PropertyPathInterface,
 *     label?: string|false|null,
 *     link_parameters?: array<string, mixed>,
 *     role?: string|string[],
 *     sort_field_mapping?: array<string, mixed>,
 *     sort_parent_association_mappings?: array<array<string, mixed>>,
 *     sortable?: string|bool,
 *     template?: string,
 *     translation_domain?: string|false,
 *     type?: string,
 *     virtual_field?: bool
 * }&array<string, mixed>
 * @phpstan-type FieldDescriptionOptions = array<string, mixed>
```

We will have trouble with PHPStan or Psalm because of that?

jordisala1991: The fieldDescrpition interface is on AdminBundle, and I don't think it makes sense to move it...

src/Model/BaseBlock.php
```diff
@@ -199,6 +199,26 @@ public function addChildren(BlockInterface $children): void
         $children->setParent($this);
     }
 
+    public function removeChild(BlockInterface $child): void
```
jordisala1991, line 202: When I made the collection inline, there was an exception asking me to add this removeChild:

```
Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException:
The add method "addChild" in class "App\Entity\SonataPageBlock" was found, but the corresponding remove method "removeChild" was not found.

  at vendor/symfony/property-access/PropertyAccessor.php:534
  at Symfony\Component\PropertyAccess\PropertyAccessor->writeProperty()
     (vendor/symfony/property-access/PropertyAccessor.php:136)
  at Symfony\Component\PropertyAccess\PropertyAccessor->setValue()
     (vendor/sonata-project/admin-bundle/src/Manipulator/ObjectManipulator.php:50)
  at Sonata\AdminBundle\Manipulator\ObjectManipulator::addInstance()
     (vendor/sonata-project/admin-bundle/src/Admin/AdminHelper.php:159)
  at Sonata\AdminBundle\Admin\AdminHelper->appendFormFieldElement()
     (vendor/sonata-project/admin-bundle/src/Action/AppendFormFieldElementAction.php:74)
  at Sonata\AdminBundle\Action\AppendFormFieldElementAction->__invoke()
     (vendor/symfony/http-kernel/HttpKernel.php:152)
  at Symfony\Component\HttpKernel\HttpKernel->handleRaw()
     (vendor/symfony/http-kernel/HttpKernel.php:74)
  at Symfony\Component\HttpKernel\HttpKernel->handle()
     (vendor/symfony/http-kernel/Kernel.php:202)
  at Symfony\Component\HttpKernel\Kernel->handle()
     (vendor/symfony/runtime/Runner/Symfony/HttpKernelRunner.php:35)
  at Symfony\Component\Runtime\Runner\Symfony\HttpKernelRunner->run()
     (vendor/autoload_runtime.php:29)
  at require_once('/usr/app/vendor/autoload_runtime.php')
     (public/index.php:7)    
```

At the following route: `https://localhost:8443/backoffice/core/append-form-field-element?_sonata_admin=sonata.page.admin.page%7Csonata.page.admin.block&_sonata_name=admin_app_sonatapagepage_sonatapageblock_edit&childId=37&elementId=s62fb36aab7a56_children&id=106&objectId=37&uniqid=s62fb36aab7a56`

I guess its caused because we added `addChild` replacing `addChildren` (before it wasn't even used that method I guess.

src/Form/Mapper/FormMapper.php
VincentLanglet: We have the Admin FormMapper, extending the Block FormMapper
https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Form/FormMapper.php#L36

FieldDescription seems to me something only related to the Admin FormMapper. Why do you need to add a 4th parameter here ?
Can't we check it's an AdminFormMapper on PageBundle ?

jordisala1991: I need to add because I am using it here: https://github.com/sonata-project/SonataBlockBundle/pull/1100/files/bde0b23549e42018ceebf8f3b064dbcfb6f607e4#diff-0dc8b125bb9655cf156cd39bc2c8e992145d4337a2c3d0cd3a195b0661322e73

And PHPStan complains.

jordisala1991: And I can't check for that on PageBundle, since the block service is placed here, and here we don't depend on AdminBundle.

IMO, if we don't want to solve it this way, another option is to deprecate this block (remove it on 5.0) and implement it directly on PageBundle.

But maybe there are other options.

VincentLanglet: > I need to add because I am using it here: https://github.com/sonata-project/SonataBlockBundle/pull/1100/files/bde0b23549e42018ceebf8f3b064dbcfb6f607e4#diff-0dc8b125bb9655cf156cd39bc2c8e992145d4337a2c3d0cd3a195b0661322e73

I see. I think the issue is kinda in the ContainerService then.
This is not really meant to be used by something else than an admin if we start to add `'edit' => 'inline'` etc, because these options are supported by SonataAdmin.

The idea of having these BlockService here should be to be able to use BlockBundle without AdminBundle.
If some blocks are too much related they should be in AdminBundle or a SpecificBundle.

SeoBundle have specific services https://github.com/sonata-project/SonataSeoBundle/tree/3.x/src/Block/Breadcrumb

So I start to wondering if the ContainerService shouldn't be move to PageBundle then.
If it doesn't work without adding `edit => inline`, that would mean that nobody use it atm.

jordisala1991: The only thing I didn't tried is if the normal collection type with the modal works at all, I will try that one first, before deprecating this.

VincentLanglet: > IMO, if we don't want to solve it this way, another option is to deprecate this block (remove it on 5.0) and implement it directly on PageBundle.

Thats exactly what I have in mind

jordisala1991: Another thing to take into account is that I am not sure if CollectionType is really useful outside SonataAdmin and the concept of Admin and entity classes:

1. https://github.com/sonata-project/form-extensions/blob/1.x/src/Type/CollectionType.php Seems really incomplete. It can only render textTypes without SonataAdmin?
2. When you use SonataAdmin, that CollectionType is related to the admin of the class you are doing collection, which we don't have here really (it fails if you don't have an admin)

So probably the best option is to deprecate this one, since it can't be used... but I will do some testing first.

VincentLanglet: > Another thing to take into account is that I am not sure if CollectionType is really useful outside SonataAdmin and the concept of Admin and entity classes:
> 
> 1. https://github.com/sonata-project/form-extensions/blob/1.x/src/Type/CollectionType.php Seems really incomplete. It can only render textTypes without SonataAdmin?
> 2. When you use SonataAdmin, that CollectionType is related to the admin of the class you are doing collection, which we don't have here really (it fails if you don't have an admin)
> 
> So probably the best option is to deprecate this one, since it can't be used... but I will do some testing first.

If we dig more, there is something wrong with the doc https://docs.sonata-project.org/projects/form-extensions/en/1.x/reference/form_types/ because all the example are given in a SonataAdmin context.

And some template are in SonataAdmin https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Resources/views/Form/form_admin_fields.html.twig#L430-L444

I dunno if all the form in form-extensions can be used correctly without SonataAdmin.

jordisala1991: Some of them should be fine, like the date picker, but this one I am not sure it makes sense 😅

src/Block/Service/ContainerBlockService.php
jordisala1991:
```suggestion
        $form->add('children', CollectionType::class);
```

src/Block/Service/ContainerBlockService.php
```diff
@@ -61,7 +61,11 @@ public function configureEditForm(FormMapper $form, BlockInterface $block): void
...
+            'edit' => 'inline',
```
jordisala1991:
```suggestion
```

src/Block/Service/ContainerBlockService.php
```diff
@@ -61,7 +61,11 @@ public function configureEditForm(FormMapper $form, BlockInterface $block): void
...
+            'inline' => 'table',
```
jordisala1991:
```suggestion
```

src/Block/Service/ContainerBlockService.php
```diff
@@ -61,7 +61,11 @@ public function configureEditForm(FormMapper $form, BlockInterface $block): void
...
+            'sortable' => 'position',
```
jordisala1991:
```suggestion
```

src/Block/Service/ContainerBlockService.php
```diff
@@ -61,7 +61,11 @@ public function configureEditForm(FormMapper $form, BlockInterface $block): void
...
+        ]);
```
jordisala1991:
```suggestion
```

src/Form/Mapper/FormMapper.php
```diff
@@ -41,10 +41,11 @@ public function reorder(array $keys);
     /**
      * @param class-string<FormTypeInterface>|null $type
      * @param array<string, mixed>                 $options
+     * @param array<string, mixed>                 $fieldDescriptionOptions
```
jordisala1991:
```suggestion
```

src/Form/Mapper/FormMapper.php
jordisala1991:
```suggestion
    public function add(string $name, ?string $type = null, array $options = []);
```

jordisala1991: Reverted this once

src/Block/Service/ContainerBlockService.php
jordisala1991: Reverted , this should be applied on SonataPageBundle.

jordisala1991:
According to our conversation:

1. this pr only adds the removechild (needed for the rest of the changes and also for consistency with addChild)
2. we have to do another pr to deprecate this block (since it can only be used on the SonataAdmin context)
3. we have to do another pr on SonataPageBundle to stop xtend this block and define all options there, included a exception is the block edit is tried to use outside sonata.
4. we might want to test this block on sonata page with panther test case (it could be followed up with tests for compose page, to test if js works correctly)

VincentLanglet:
> According to our conversation:
> 
> 1. this pr only adds the removechild (needed for the rest of the changes and also for consistency with addChild)
> 2. we have to do another pr to deprecate this block (since it can only be used on the SonataAdmin context)
> 3. we have to do another pr on SonataPageBundle to stop xtend this block and define all options there, included a exception is the block edit is tried to use outside sonata.
> 4. we might want to test this block on sonata page with panther test case (it could be followed up with tests for compose page, to test if js works correctly)

Changelog need to be updated

VincentLanglet (approved):

VincentLanglet:
> 2. we have to do another pr to deprecate this block (since it can only be used on the SonataAdmin context)
> 3. we have to do another pr on SonataPageBundle to stop xtend this block and define all options there, included a exception is the block edit is tried to use outside sonata.
> 4. we might want to test this block on sonata page with panther test case (it could be followed up with tests for compose page, to test if js works correctly)

Do you plan to do it or should we open issues ?

jordisala1991:
Yes, I plan to do it.

Issue #1109, opened 2022-09-12
Deprecate ContainerBlockService on 4.x


jordisala1991:
## Feature Request

That block make no sense outside AdminBundle, it needs to be deprecated and removed before 5.0

VincentLanglet:
> That block make no sense outside AdminBundle, it needs to be deprecated and removed before 5.0

And moved to another Bundle then or just deprecated/removed ?

jordisala1991:
it was already moved to PageBundle, here we just need deprecate + removal

VincentLanglet:
I think I can close this seems it's not sure anymore, cf discussion on https://github.com/sonata-project/SonataBlockBundle/pull/1114

PR #1112, opened 2022-09-17
POC to reuse MenuBlock

VincentLanglet:
## Changelog

```markdown
### Added
- AbstractMenuBlockService

### Deprecated
- MenuBlockService current_uri option
- MenuBlockService menu_class option
- MenuBlockService children_class option
```

### DIFF
--- a/src/Block/Service/AbstractMenuBlockService.php
+++ b/src/Block/Service/AbstractMenuBlockService.php
@@ -16,8 +16,11 @@ namespace Sonata\BlockBundle\Block\Service;
 use Knp\Menu\ItemInterface;
 use Sonata\BlockBundle\Block\BlockContextInterface;
 use Sonata\BlockBundle\Form\Mapper\FormMapper;
+use Sonata\BlockBundle\Meta\Metadata;
+use Sonata\BlockBundle\Meta\MetadataInterface;
 use Sonata\BlockBundle\Model\BlockInterface;
 use Sonata\Form\Type\ImmutableArrayType;
+use Sonata\Form\Validator\ErrorElement;
 use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
@@ -26,7 +29,7 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Twig\Environment;
 
-abstract class AbstractMenuBlockService extends AbstractBlockService
+abstract class AbstractMenuBlockService extends AbstractBlockService implements EditableBlockService
 {
     public function __construct(Environment $twig)
     {
@@ -66,6 +69,20 @@ abstract class AbstractMenuBlockService extends AbstractBlockService
         ]);
     }
 
+    public function validate(ErrorElement $errorElement, BlockInterface $block): void
+    {
+    }
+
+    /**
+     * NEXT_MAJOR: Remove this method.
+     */
+    public function getMetadata(): MetadataInterface
+    {
+        return new Metadata('sonata.block.service.menu', null, null, 'SonataBlockBundle', [
+            'class' => 'fa fa-bars',
+        ]);
+    }
+
     public function configureSettings(OptionsResolver $resolver): void
     {
         $resolver->setDefaults([
@@ -77,8 +94,11 @@ abstract class AbstractMenuBlockService extends AbstractBlockService
             'current_class' => 'active',
             'first_class' => false,
             'last_class' => false,
+            // NEXT_MAJOR: Remove.
             'current_uri' => null,
+            // NEXT_MAJOR: Remove.
             'menu_class' => 'list-group',
+            // NEXT_MAJOR: Remove.
             'children_class' => 'list-group-item',
             'menu_template' => null,
         ]);
@@ -89,6 +109,33 @@ abstract class AbstractMenuBlockService extends AbstractBlockService
             ...$this->deprecationParameters(
                 '4.12',
                 'Option "cache_policy" is deprecated since sonata-project/block-bundle 4.12 and will be removed in 5.0.'
+            )
+        );
+
+        // NEXT_MAJOR: Remove setDeprecated.
+        $resolver->setDeprecated(
+            'current_uri',
+            ...$this->deprecationParameters(
+                '4.x',
+                'Option "current_uri" is deprecated since sonata-project/block-bundle 4.x and will be removed in 5.0.'
+            )
+        );
+
+        // NEXT_MAJOR: Remove setDeprecated.
+        $resolver->setDeprecated(
+            'menu_class',
+            ...$this->deprecationParameters(
+                '4.x',
+                'Option "menu_class" is deprecated since sonata-project/block-bundle 4.x and will be removed in 5.0.'
+            )
+        );
+
+        // NEXT_MAJOR: Remove setDeprecated.
+        $resolver->setDeprecated(
+            'children_class',
+            ...$this->deprecationParameters(
+                '4.x',
+                'Option "children_class" is deprecated since sonata-project/block-bundle 4.x and will be removed in 5.0.'
             )
         );
     }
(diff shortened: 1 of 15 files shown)


src/Block/Service/AbstractMenuBlockService.php
```diff
@@ -0,0 +1,197 @@
...
+        $resolver->setDefaults([
+            'title' => '',
+            // NEXT_MAJOR: Remove.
+            'cache_policy' => 'public',
+            'template' => '@SonataBlock/Block/block_core_menu.html.twig',
+            'safe_labels' => false,
+            'current_class' => 'active',
+            'first_class' => false,
+            'last_class' => false,
+            'current_uri' => null,
```
VincentLanglet: Current_uri doesn't seems to be used here, but only in SeoBundle. I propose to remove it (in next major).

src/Block/Service/AbstractMenuBlockService.php
```diff
@@ -0,0 +1,246 @@
...
+            // NEXT_MAJOR: Remove.
+            'menu_class' => 'list-group',
+            // NEXT_MAJOR: Remove.
+            'children_class' => 'list-group-item',
+            'menu_template' => null,
+        ]);
+
+        // NEXT_MAJOR: Remove setDeprecated.
+        $resolver->setDeprecated(
```
core23, line 83: This looks wrong. The class is brand new and introduces deprecations?

Can you move the options and deprecations to the using class.

src/Block/Service/AbstractMenuBlockService.php
```diff
@@ -0,0 +1,246 @@
...
+            'translation_domain' => 'SonataBlockBundle',
+        ]);
+    }
+
+    public function validate(ErrorElement $errorElement, BlockInterface $block): void
+    {
+    }
+
+    /**
+     * NEXT_MAJOR: Remove this method.
```
VincentLanglet: This will require this: https://github.com/sonata-project/SonataBlockBundle/pull/1113 if it's fine for you @jordisala1991 @core23

src/Block/Service/MenuBlockService.php
```diff
@@ -130,100 +105,41 @@ public function getMetadata(): MetadataInterface
...
-     * @return array<array{string, class-string<FormTypeInterface>, array<string, mixed>}>
-     */
-    private function getFormSettingsKeys(): array
+    protected function getFormSettingsKeys(): array
     {
         $choiceOptions = [
             'required' => false,
-            'label' => 'form.label_url',
+            'label' => 'form.label_menu_name',
             'choice_translation_domain' => 'SonataBlockBundle',
```
VincentLanglet: I'm not sure the choice_translation_domain must be SonataBlockBundle here, since the alias name are coming from `$this->menuRegistry->getAliasNames()` so this are alias from the user. (cc @jordisala1991  @core23 )

src/Block/Service/MenuBlockService.php
```diff
@@ -130,100 +105,41 @@ public function getMetadata(): MetadataInterface
...
 
-    /**
-     * @return array<array{string, class-string<FormTypeInterface>, array<string, mixed>}>
-     */
-    private function getFormSettingsKeys(): array
+    protected function getFormSettingsKeys(): array
     {
         $choiceOptions = [
             'required' => false,
-            'label' => 'form.label_url',
```
VincentLanglet, line 142: form.label_url was a bad copy paste, since it display "URL" when it's was the menu name instead.

The fact that nobody opened an issue/PR about it would means that nobody use this class.

VincentLanglet: I start wondering if the MenuBlockService is use by someone then,

If not, we could follow the same strategy than https://github.com/sonata-project/SonataBlockBundle/issues/1109 and deprecate the MenuBlock and move all this code to the SeoBundle.
Everything would be in the BaseBreadcrumbService and this would solve all the issues...

src/Block/Service/AbstractMenuBlockService.php
jordisala1991: Can we separate those changes? I am not really sure TBH.

src/Block/Service/MenuBlockService.php
jordisala1991: I would say let's focus on the BC Break fix, I would not touch the topic of deprecating all of this.

src/Block/Service/AbstractMenuBlockService.php
jordisala1991: Does this make sense to you @VincentLanglet ?

src/Block/Service/MenuBlockService.php
jordisala1991: I would say to touch this topic on another PR if any. I guess this has been like this for ages.

jordisala1991 (dismissed):
I would say to revert some deprecations but in general 👍

src/Block/Service/AbstractMenuBlockService.php
VincentLanglet: The issue is the fact you need to provide a Metadata so far, and a metadata name then.

Does it make sens to provide a Metadata in an abstract class ? This would mean that every block extending it will have the same. (Unless they override it).

jordisala1991: You don't have to, just declare the method abstract and implement in the next class.

VincentLanglet: That's what I wanted to avoid because of the whole BC-break topic https://github.com/sonata-project/SonataSeoBundle/pull/691.

This means that `getMetadata` will become an abstract method of https://github.com/sonata-project/SonataSeoBundle/pull/692/files#diff-92b90700009daba6f28946d432d0739399d3d9b55ea929e882923368a029b46f so user extending it will have to implements it and that's a BC break.

The other solution would be to implements it in SonataSeoBundle, with a NEXT_MAJOR: remove the implementation.
This way there is no BC break, and people will have to implement it in next major.

jordisala1991: i see 3 options:

1. declare the method normally here and make it abstract on 5.x (not sure how good the idea will be in terms of supporting multiple block bundle versions) to avoid BC break
2. split the abstract block in a class + trait (like we discussed on slack) to avoid BC break
3. Assume the bc break.

jordisala1991: In terms of avoiding BC breaks, imo the 2nd is the best, because it will allow to not have an editable block for breadcrumbs.

VincentLanglet: I don't really like trait so wanted to avoid solution 2 and i'm more in favor of the solution 1.
This seems ok for the support of multiple block bundle version, you can override the method.

But still I wonder if the split of the interface is not a good idea. (but can be later)
You might edit those block without the need of a Metadata.

VincentLanglet:
This should be ready for review @jordisala1991

src/Block/Service/AbstractMenuBlockService.php
```diff
@@ -0,0 +1,196 @@
...
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\Form\Extension\Core\Type\TextType;
+use Symfony\Component\Form\FormTypeInterface;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Twig\Environment;
+
+abstract class AbstractMenuBlockService extends AbstractBlockService implements EditableBlockService
+{
+    public function __construct(Environment $twig)
```
jordisala1991: Do we need this construct?

VincentLanglet: Removed

src/Block/Service/AbstractMenuBlockService.php
```diff
@@ -0,0 +1,191 @@
...
+use Sonata\Form\Validator\ErrorElement;
+use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\Form\Extension\Core\Type\TextType;
+use Symfony\Component\Form\FormTypeInterface;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Twig\Environment;
+
+abstract class AbstractMenuBlockService extends AbstractBlockService implements EditableBlockService
```
jordisala1991, line 29: At the end you choose to not add the metadata method here then? will it be on the breadcrumb classes then?

VincentLanglet: This class is new, so I'm able to not add the metadata method here and let user implement it.

In the seo PR https://github.com/sonata-project/SonataSeoBundle/pull/692 I provided a default implementation to avoid the BC break, and we'll remove the default implementation in next major.

I still wonder if it wouldn't a good idea to split editableBlock but that can be another topic :p

jordisala1991 (dismissed):

jordisala1991:
cs fixer failing

VincentLanglet:
> cs fixer failing

Solved

jordisala1991 (approved):

PR #1113, opened 2022-09-17
Split EditableBlockService interface

VincentLanglet:
## Subject

I might be wrong but the `getMetadata` seems to be only used in PageBundle, and I'm not sure if it should really be related to the EditableBlockServiceInterface. IMHO you can edit something without metadata. And you can have something with metadata which is not editable.

## Changelog

```markdown
### Added
- MetadataAwareBlockServiceInterface

### Deprecated
- EditableServiceInterface::getMetadata
```

### DIFF
--- a/src/Block/Service/ContainerBlockService.php
+++ b/src/Block/Service/ContainerBlockService.php
@@ -32,7 +32,7 @@ use Symfony\Component\OptionsResolver\OptionsResolver;
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
  */
-final class ContainerBlockService extends AbstractBlockService implements EditableBlockService
+final class ContainerBlockService extends AbstractBlockService implements MetadataAwareBlockServiceInterface, EditableBlockService
 {
     public function configureCreateForm(FormMapper $form, BlockInterface $block): void
     {
--- a/src/Block/Service/EditableBlockService.php
+++ b/src/Block/Service/EditableBlockService.php
@@ -14,20 +14,19 @@ declare(strict_types=1);
 namespace Sonata\BlockBundle\Block\Service;
 
 use Sonata\BlockBundle\Form\Mapper\FormMapper;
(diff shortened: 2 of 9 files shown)


jordisala1991 (commented):
Splitting this class will require more work in the page bundle. Right now you can kind of know if a block is editable if it implements the EditableBlockService, with your change it won't be possible anymore.

We will need to check for EditableBlockService to know if there is a form to be used, AND for MetadataAware to be ABLE to print the block name and provide a link to edit form. I don't think it is a good idea.

VincentLanglet:
> Splitting this class will require more work in the page bundle. Right now you can kind of know if a block is editable if it implements the EditableBlockService, with your change it won't be possible anymore

I thought this was a good idea, because:
- PageBundle is the only one using things related to MetadataAware interface.
- of this https://github.com/sonata-project/SonataBlockBundle/pull/1112/files#r979612860 to avoid providing a Metadata in an abstract class because I thought this had no sens.

If having multiples blocks with the same metadata (and metadata name) is ok we can keep it like this. But the fact that in PageBundle we override the Metadata let me think it's not. I didn't want to add an abstract method because it would require to implements it in all MenuBlock which would be an issue for people not using the Editable things (like core23).

> We will need to check for EditableBlockService to know if there is a form to be used, AND for MetadataAware to be ABLE to print the block name and provide a link to edit form. I don't think it is a good idea.

When you describe things like this, it seems to be 2 different things. You can edit a form without the need to have a `getMetadata` methods. Maybe would it be better if `MetadataAwareBlockServiceInterface extends EditableBlockService` wdyt @jordisala1991 ?

jordisala1991:
But would that simplify anything? you will still require to implement both 3 methods on all blocks. Unless you want to create duplicate blocks to add the metadata thing.

Its the same as the form methods, they are only used in the admin, but If you want to use outside sonata to print your own form, you most likely will need some kind of meta information to be able to know how a block is named.

I think like there is a relation between be able to create/edit a block and knowing its "human" name.

PR #1115, opened 2022-09-19
Do not trigger errors when a block doesn't exist anymore

jeromeengeln:
## Keep the blocks into the page rendering even if some blocks does not exist

Here a proposition to remove exception on render when a block doesn't not exists anymore.

The purpose of changing exception is to avoid throwing an exception for blocks that does not exist anymore (for example a removed service and still present into page block composition - issue https://github.com/sonata-project/SonataPageBundle/issues/1609 in sonata page) and keep all other runtime errors.
To do that I need to identify not found exception.

if BlockNotFoundException is catched just return an empty response
And for other \RuntimeException just keep old behavior

Related to SonataPageBundle issue https://github.com/sonata-project/SonataPageBundle/issues/1609
And PR  https://github.com/sonata-project/SonataPageBundle/pull/1611

## Changelog

```markdown
### Changed
- BlockServiceManager::load throw a BlockServiceNotFoundException when the block is not found.
```


eerison:
~I was working in this issue as well 😄~

~But I don't get why you need to change those class?~

~I just needed to change PageExtension~

sorry I misunderstand your PR ;)

eerison:
can you provide tests please!

jeromeengeln:
> can you provide tests please!

you mean create a new unit test for those changes ? I'm not really familiar with that practice but if you guide me I can try

src/Block/BlockRenderer.php
```diff
@@ -79,6 +82,10 @@ public function render(BlockContextInterface $blockContext, ?Response $response
                 ), compact('exception'));
             }
 
+            if ($exception instanceof RuntimeError && !empty($exception->getPrevious()) && $exception->getPrevious() instanceof BlockNotFoundException) {
```
jordisala1991: I don't think this code belongs here. We have ExceptionStrategyManager for that I guess.

VincentLanglet: Yes this could be here instead https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Exception/Strategy/StrategyManager.php#L107-L125

which would rely on https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Exception/Strategy/StrategyManager.php#L113-L114

So maybe it's just a blockRenderer/filter to update in BlockBundle or Pagebundle

src/Block/BlockServiceManager.php
```diff
@@ -190,20 +191,18 @@ public function validate(ErrorElement $errorElement, BlockInterface $block): voi
     }
 
     /**
-     * @throws \RuntimeException
```
VincentLanglet, line 193: This shouldn't be removed

VincentLanglet:
> > can you provide tests please!
> 
> you mean create a new unit test for those changes ? I'm not really familiar with that practice but if you guide me I can try

This can be a functional test (might be better).

The current issue for me is to understand the issue you and @eerison are trying to fix.
So best would be
- A commit to reproduce the error with functional tests (which means a failing build)
- A commit with your fix (and the tests will pass)

Because currently I'm like "Why should be have a different treatment for those errors ?"

Also, looking at the code (cf https://github.com/sonata-project/SonataBlockBundle/pull/1115#discussion_r974163680) it seems like this could be already configured to not throw an exception but render some error message...

Also I don't see why it solve https://github.com/sonata-project/SonataPageBundle/issues/1609 and replace https://github.com/sonata-project/SonataPageBundle/pull/1611/files

src/Block/BlockServiceManager.php
jeromeengeln: Yeah my bad, I restored @throws initial tag and add the other exception

jeromeengeln:
> Because currently I'm like "Why should be have a different treatment for those errors ?"

Yeah the purpose of changing exception is to avoid throwing an exception for blocks that does not exist anymore (for example a removed service and still present into page block composition - issue [1609](https://github.com/sonata-project/SonataPageBundle/issues/1609) in sonata page) and keep all other runtime errors.
To do that I need to identify not found exception.
- if BlockNotFoundException is catched just return an empty response
- And for other \RuntimeException just keep old behavior

eerison:
> A commit to reproduce the error with functional tests (which means a failing build)
> A commit with your fix (and the tests will pass)

Yep!

VincentLanglet:
> > Because currently I'm like "Why should be have a different treatment for those errors ?"
> 
> Yeah the purpose of changing exception is to avoid throwing an exception for blocks that does not exist anymore (for example a removed service and still present into page block composition - issue [1609](https://github.com/sonata-project/SonataPageBundle/issues/1609) in sonata page) and keep all other runtime errors. To do that I need to identify not found exception.
> 
> * if BlockNotFoundException is catched just return an empty response
> * And for other \RuntimeException just keep old behavior

Doing a PR to throws BlockNotFoundException instead of \RuntimeException seems fine to me.
We already use BlockNotFoundException in some places.

The main question is about the way we handle BlockNotFoundException.
-> Should be handled differently in BlockBundle and why ?
-> Should be handle differently in PageBundle instead and why ?

A failing (functional) test would help to understand the why.

We have multiple renderer https://github.com/sonata-project/SonataBlockBundle/tree/4.x/src/Exception/Renderer and filter https://github.com/sonata-project/SonataBlockBundle/tree/4.x/src/Exception/Filter ; might be interesting to know which ones are used in PageBundle that would help to understand why an exception is thrown.

jeromeengeln:
> Also I don't see why it solve [sonata-project/SonataPageBundle#1609](https://github.com/sonata-project/SonataPageBundle/issues/1609) and replace https://github.com/sonata-project/SonataPageBundle/pull/1611/files

For [sonata-project/SonataPageBundle#1609](https://github.com/sonata-project/SonataPageBundle/issues/1609)
It allow to avoid the exception mentionned in issue (the front-end part) :
```
An exception has been thrown during the rendering of a template ("The block service `sonata.seo.block.email.share_button` does not exist").
```

And the PR [sonata-project/SonataPageBundle#1611](https://github.com/sonata-project/SonataPageBundle/pull/1611/files) just fix and improve the sonata page compose view behavior when some block has been removed (not throw 500 on block list and allow to remove even if the block does not exists anymore). So it's merly related in our discussions with @eerison

VincentLanglet:
> > Also I don't see why it solve [sonata-project/SonataPageBundle#1609](https://github.com/sonata-project/SonataPageBundle/issues/1609) and replace https://github.com/sonata-project/SonataPageBundle/pull/1611/files
> 
> For [sonata-project/SonataPageBundle#1609](https://github.com/sonata-project/SonataPageBundle/issues/1609) It allow to avoid the exception mentionned in issue (the front-end part) :
> 
> ```
> An exception has been thrown during the rendering of a template ("The block service `sonata.seo.block.email.share_button` does not exist").
> ```

Yes I looked at the code and kinda understand more the issue.
SonataPage is using the BlockHelper from BlockBundle to render the block which is throwing the exception.

But since there is an exceptionManagerStrategy, we can use a different one.
Or basically just add a try/catch (BlockNotFoundException) here: https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Twig/Extension/PageExtension.php#L236

So there are multiple way to solve this and we need to precisely decide if it must be fix in SonataPage or SonataBock.
I'm tempted to think that the best would be to
- Use BlockNotFoundException in BlockBundle but do not touch the way BlockHelper works
- Solve the issue in SonataPage by changing the exceptionRendererStrategy or adding a try catch if another exceptionRendererStrategy doesn't make sens.

jeromeengeln:
> The main question is about the way we handle BlockNotFoundException.
> -> Should be handled differently in BlockBundle and why ?
> -> Should be handle differently in PageBundle instead and why ?

I see the idea, I'll think about it.

eerison:
> Or basically just add a try/catch (BlockNotFoundException) here: https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Twig/Extension/PageExtension.php#L236

Yep, I'm doing exactly this now :)

eerison:
to solve the issue related with page (when open the page it's returning internal server error), I just need this line

https://github.com/sonata-project/SonataBlockBundle/pull/1115/files#diff-26110fbfbdde2b86990d5fc9dd60c1f629a5688cb6bb0932f0f9cdd5bbfb4de9R200

eerison:
> to solve the issue related with page (when open the page it's returning internal server error), I just need this line
> 
> https://github.com/sonata-project/SonataBlockBundle/pull/1115/files#diff-26110fbfbdde2b86990d5fc9dd60c1f629a5688cb6bb0932f0f9cdd5bbfb4de9R200

~But should be good~ We must write a test for it!

VincentLanglet:
> to solve the issue related with page (when open the page it's returning internal server error), I just need this line
> 
> https://github.com/sonata-project/SonataBlockBundle/pull/1115/files#diff-26110fbfbdde2b86990d5fc9dd60c1f629a5688cb6bb0932f0f9cdd5bbfb4de9R200

Yeah, I think this PR should be reduce to just changing some \RuntimeException to BlockNotFoundException
Wdyt @jordisala1991 @JeromeEngelnAdeliom ?

jeromeengeln:
> > to solve the issue related with page (when open the page it's returning internal server error), I just need this line
> > https://github.com/sonata-project/SonataBlockBundle/pull/1115/files#diff-26110fbfbdde2b86990d5fc9dd60c1f629a5688cb6bb0932f0f9cdd5bbfb4de9R200
> 
> Yeah, I think this PR should be reduce to just changing some \RuntimeException to BlockNotFoundException Wdyt @jordisala1991 @JeromeEngelnAdeliom ?

Fine for me ! I did the rollback commit

src/Block/BlockServiceManager.php
```diff
@@ -191,19 +192,19 @@ public function validate(ErrorElement $errorElement, BlockInterface $block): voi
 
     /**
      * @throws \RuntimeException
+     * @throws BlockNotFoundException
      */
     private function load(string $type): BlockServiceInterface
     {
         if (!$this->has($type)) {
-            throw new \RuntimeException(sprintf('The block service `%s` does not exist', $type));
+            throw new BlockNotFoundException(sprintf('The block service `%s` does not exist', $type));
```
eerison: A unit test is required hire!

src/Block/BlockServiceManager.php
```diff
@@ -191,19 +192,19 @@ public function validate(ErrorElement $errorElement, BlockInterface $block): voi
...
         }
 
         if (!$this->services[$type] instanceof BlockServiceInterface) {
             $blockService = $this->container->get($type);
             if (!$blockService instanceof BlockServiceInterface) {
                 throw new \RuntimeException(sprintf('The service %s does not implement BlockServiceInterface', $type));
             }
-
```
eerison, line 206: revert this change please!

eerison (changes requested):

src/Block/BlockServiceManager.php
jeromeengeln: Yeah, I'll try to do it

VincentLanglet (dismissed):

src/Block/BlockServiceManager.php
jeromeengeln: I added a test, If you have some comments to improve do not hesitate or the test is not well organized

tests/Exception/Renderer/BlockNotFoundRendererTest.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        $this->expectException(BlockNotFoundException::class);
+
+        $service = $this->createMock(BlockServiceInterface::class);
+
+        $container = new Container();
+
+        $container->set('existing-block', $service);
+
+        $manager = new BlockServiceManager($container, []);
```
VincentLanglet: You're testing the BlockServiceManager so the test should be in https://github.com/sonata-project/SonataBlockBundle/blob/4.x/tests/Block/BlockServiceManagerTest.php

I think this is already tested and you just need to update https://github.com/sonata-project/SonataBlockBundle/blob/4.x/tests/Block/BlockServiceManagerTest.php#L59-L61

VincentLanglet (changes requested):

tests/Exception/Renderer/BlockNotFoundRendererTest.php
jeromeengeln: Oh I see, I'll do that, thank you

eerison:
Hey @VincentLanglet  will it be a minor or patch version?

eerison (dismissed):

VincentLanglet:
> Hey @VincentLanglet will it be a minor or patch version?

Minor, it's not a fix but sort of a new feature

VincentLanglet (dismissed):

src/Block/BlockServiceManager.php
```diff
@@ -191,11 +192,12 @@ public function validate(ErrorElement $errorElement, BlockInterface $block): voi
 
     /**
      * @throws \RuntimeException
+     * @throws BlockNotFoundException
      */
     private function load(string $type): BlockServiceInterface
     {
         if (!$this->has($type)) {
-            throw new \RuntimeException(sprintf('The block service `%s` does not exist', $type));
+            throw new BlockNotFoundException(sprintf('The block service `%s` does not exist', $type));
```
jordisala1991: This is not really correct right? Block is found, what is not found is the BlockService, maybe it needs a new Exception class? like BlockServiceNotFound?

eerison: Hmmmm I thought that block and block service are the same, but if they are different, then we need to create a new exception, because we are returning a block service

VincentLanglet: Yes indeed, might be great to introduce BlockServiceNotFoundException

jeromeengeln: Done in my last commit, Wdyt ?

src/Exception/BlockServiceNotFoundException.php
```diff
@@ -0,0 +1,20 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\BlockBundle\Exception;
+
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+final class BlockServiceNotFoundException extends NotFoundHttpException
+{
```
eerison, line 17: Should it extend from http exception 🤔

If yes then we should add NotFoundHttpException as Suffix?!

jeromeengeln: Maybe not, I copied the old exception. And what about that ?
```php
namespace Sonata\BlockBundle\Exception;

final class BlockServiceNotFoundException implements BlockExceptionInterface
{
}
```
Or simply :
```php
final class BlockServiceNotFoundException
{
```
Let me know and I'll do modifications tomorrow morning

jordisala1991: it needs to extend some kind of exception. If you also implement the BlockExceptionInterface maybe thats enough to render empty response automatically? or not?

I feel like the whole exception catching should happen in the block bundle, are you sure you want to introduce specific logic for page @VincentLanglet ?

eerison: > I feel like the whole exception catching should happen in the block bundle, are you sure you want to introduce specific logic for page @VincentLanglet ?

Hey @jordisala1991, do you mean here https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L388  ?

But if we do this, we can't handle the exception for code that uses this method, if for some reason I need to implement some logical if the block doesn't exist, I can't do that, because I never get the exception!

usually I handle the exception in the last layer before to return to client.

VincentLanglet: First I dunno why BlockNotFound extends NotFoundException, it feels weird https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Exception/BlockNotFoundException.php#L18

I mean, I don't see how we're in a Http context here: https://github.com/sonata-project/SonataBlockBundle/blob/375b168da31545bf3c386fd25a4b483411ac97c5/src/Block/BlockLoaderChain.php#L66 and we should throw a 404.

And indeed there is a BlockExceptionInterface.

Seems like a strategy exist to ignore block exceptions https://github.com/sonata-project/SonataBlockBundle/blob/987d2936d3d2a3d4fe4615241f545aad693eb4c2/docs/reference/exceptions.rst#filters

So indeed, throwing a BlockServiceNotFound extends BlockExceptionInterface would allow people to ignore this exception with the correct filter.
It might require some conf in PageBundle still to use the correct filter.

But again, I asked a question and got no answer so far:
which filter and which renderer is using SonataPageBundle.
https://github.com/sonata-project/SonataBlockBundle/blob/987d2936d3d2a3d4fe4615241f545aad693eb4c2/docs/reference/exceptions.rst#exception-strategy

IMHO, blockbundle should allow to ignore them (with specific exception/config) and pagebundle should activate this ; and this already a feature from blockbundle.

eerison: I didn't know this exception strategy in block bundle 👀 
Well if we return an exception that implements this interface we just need to enable it into the configs! we can  let this handle as true for default block config :) (into the page bundle docs)

eerison: we have many ways to handle it 😄

jeromeengeln: So to go forward I have to remove old `extend NotFoundHttpException` and add `implements BlockExceptionInterface ` in both files src/Exception/BlockServiceNotFoundException.php and src/Exception/BlockNotFoundException.php ?

```php
namespace Sonata\BlockBundle\Exception;
```

```php
final class BlockServiceNotFoundException implements BlockExceptionInterface
{
}
```
```php
final class BlockNotFoundException implements BlockExceptionInterface
{
}
```

And change configs into page bundle with [exception strategy feature](https://github.com/sonata-project/SonataBlockBundle/blob/987d2936d3d2a3d4fe4615241f545aad693eb4c2/docs/reference/exceptions.rst#exception-strategy). @eerison do you look for this part ?

VincentLanglet: I'm not sure for BlockNotFoundException, technically it would be a BC break

jeromeengeln: So:
- no changes for `BlockNotFoundException.php`
- and only add implements `BlockExceptionInterface` for `BlockServiceNotFoundException.php`

eerison: > I'm not sure for BlockNotFoundException, technically it would be a BC break

It's a good example to change just what is really necessary 😄

eerison: @JeromeEngelnAdeliom  could you change only `BlockServiceNotFoundException` please!

and other thing I guess you need to extend from some exception, right @VincentLanglet ?

in this case which exception? `Exception` or `NotFoundHttpException`?

jordisala1991: It would also be nice if someone checks why that interface was introduced and why they didn't implement any exception with that interface, maybe there is a reason, maybe its just one more "WTF".

eerison: After @JeromeEngelnAdeliom  change it, I'll test in sonata page bundle, and check if those configs works as expected :D

VincentLanglet: > It would also be nice if someone checks why that interface was introduced and why they didn't implement any exception with that interface, maybe there is a reason, maybe its just one more "WTF".

Ten years ago https://github.com/sonata-project/SonataBlockBundle/pull/36 :(

jeromeengeln: @eerison  wdyt about using `extends \UnexpectedValueException` ? it seems to match with our missing block service case.
=> Exception thrown if a value does not match with a set of values. Typically this happens when a function calls another function and expects the return value to be of a certain type or value not including arithmetic or buffer related errors.

eerison: IMO `BlockServiceNotFoundException` isn't an abstraction of `UnexpectedValueException` 👀

VincentLanglet: Extending Exception is enough.

And fun fact, @jordisala1991, https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Exception/BlockExceptionInterface.php#L19 is currently not extending Throwable...

eerison:
Hey @VincentLanglet  I added this config

```yaml
sonata_block:
    exception:
        default:
            filter: debug_only
        filters:
            debug_only: sonata.block.exception.filter.debug_only
            ignore_block_exception: sonata.block.exception.filter.ignore_block_exception
            keep_all: sonata.block.exception.filter.keep_all
            keep_none: sonata.block.exception.filter.keep_none
```

But it didn't handled the error

Screenshot 2022-09-20 at 13 04 20

and if I set the `APP_DEBUG=false` any block is rendered

Screenshot 2022-09-20 at 13 03 38

am I missing something?

VincentLanglet:
> am I missing something?

Be aware that I strictly no nothing about this ; just looking into the doc.
I would say with `filter: debug_only`, you're not using the `ignore_block_exception` then.

Also, might be an issue to require that the developper have to add some sonata_block config...

eerison:
> > am I missing something?
> 
> Be aware that I strictly no nothing about this ; just looking into the doc. I would say with `filter: debug_only`, you're not using the `ignore_block_exception` then.
> 
> Also, might be an issue to require that the developper have to add some sonata_block config...

Ok, it's not working, because the error comes from `object(Twig\Error\RuntimeError)[1085]`, and as it doesn't extend from BlockInterfaceException, it isn't catch into the `strategyManager`

I liked the strategy exception you can handle the logs messages in one place only, quite good!
But we need to investigate why the error comes from twig :/

eerison:
This strategy is used here (only): https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockRenderer.php#L85

and what is passed to there is the container block, and not the children blocks

```
object(App\Entity\SonataPageBlock)[524]
  protected 'name' => string 'Main content' (length=12)
  protected 'settings' => 
    array (size=1)
      'code' => string 'content' (length=7)
  protected 'enabled' => boolean true
  protected 'position' => int 1
  protected 'parent' => null
  protected 'children' => 
    array (size=3)
      0 => 
        object(App\Entity\SonataPageBlock)[463]
          protected 'name' => string 'email 1' (length=7)
          protected 'settings' => 
            array (size=3)
              ...
          protected 'enabled' => boolean true
          protected 'position' => int 1
          protected 'parent' => 
            &object(App\Entity\SonataPageBlock)[524]
          protected 'children' => 
            array (size=0)
              ...
          protected 'createdAt' => 
            object(DateTime)[456]
              ...
          protected 'updatedAt' => 
            object(DateTime)[459]
              ...
          protected 'type' => string 'sonata.seo.block.email.share_button' (length=35)
          protected 'ttl' => int 86400
          protected 'id' => int 20
          protected ?Sonata\PageBundle\Model\PageInterface 'page' => 
            object(Sonata\PageBundle\Model\SnapshotPageProxy)[641]
              ...
      1 => 
```

for this reason when the exception/block is skipped, the page is empty, because the whole snapshot is skipped.

eerison:
well I guess this PR can be merged, because for both solution (strategy or handle directly in sonata page) I need this new exception :)

eerison (dismissed):

eerison:
I invested a considerable time, to use the strategy log with page bundle

But the Big issue is, I'm passing the `container block` and the exception is related with a children block, and the whole container is skipped :/ 

Screenshot 2022-09-21 at 11 48 35

I'm without ideas :(

I would say that the current implementation for log's strategy is working as planned, it's skipping the block the contain issue, But we want to validate this for children

jeromeengeln:
> I'm without ideas

And modify [IgnoreClassFilter](https://github.com/sonata-project/SonataBlockBundle/blob/375b168da31545bf3c386fd25a4b483411ac97c5/src/Exception/Filter/IgnoreClassFilter.php#L35) used to filter `ignore_block_exception` to check into previous exception ?
```php
return !$exception instanceof $this->class;
```
=>
```php
return !$exception instanceof $this->class && !(($exception->getPrevious() ?? null) instanceof $this->class);
```

eerison:
> > I'm without ideas
> 
> And modify [IgnoreClassFilter](https://github.com/sonata-project/SonataBlockBundle/blob/375b168da31545bf3c386fd25a4b483411ac97c5/src/Exception/Filter/IgnoreClassFilter.php#L35) used to filter `ignore_block_exception` to check into previous exception ?
> 
> ```
> return !$exception instanceof $this->class;
> ```
> 
> =>
> 
> ```
> return !$exception instanceof $this->class && !(($exception->getPrevious() ?? null) instanceof $this->class);
> ```

I tried it, But it doesn't work

always it throws TwigException

eerison:
is it missing something to merge this PR?

VincentLanglet:
> is it missing something to merge this PR?

To be sure it does what we want.

For the issue you encounter with the strategy manager doing nothing, I would say it's because
- The strategyManager is used by the BlockRendered inside this call https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L451
- You currently get the exception from this line https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L412

Can you confirm this @eerison  ?

eerison:
> > is it missing something to merge this PR?
> 
> To be sure it does what we want.
> 
> For the issue you encounter with the strategy manager doing nothing, I would say it's because
> 
> * The strategyManager is used by the BlockRendered inside this call https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L451
> * You currently get the exception from this line https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L412
> 
> Can you confirm this @eerison ?

Ok Now I got a better overview why exception strategy is not working in my case!

I'm trying to get the block that doesn't exist, exactly from this line: https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L390

and as we don't have any try/catch there like here https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockRenderer.php#L85

it's crashing my page!

I did just a simple test like this

```php
        try {
            $blockContext = $this->blockContextManager->get($block, $options);
        } catch (\Throwable $exception) {
            return '';
        }
```

and it works!

The question is, should we handle strategy exception there as well?

@VincentLanglet

eerison:
If we do this here, I don't need to change anything in page bundle

VincentLanglet:
@eerison, when you tried
```
try {
            $blockContext = $this->blockContextManager->get($block, $options);
        } catch (\Throwable $exception) {
            return '';
        }
```
what was the exception thrown ?

Because looking at the code 
`blockContextManager->get` is supposed to thrown BlockNotFoundException already, so this wouldn't require any PR from BlokcBundle if we try catch the BlockNotFound in PageBundle.

eerison:
> @eerison, when you tried
> 
> ```
> try {
>             $blockContext = $this->blockContextManager->get($block, $options);
>         } catch (\Throwable $exception) {
>             return '';
>         }
> ```
> 
> what was the exception thrown ?
> 
> Because looking at the code `blockContextManager->get` is supposed to thrown BlockNotFoundException already, so this wouldn't require any PR from BlokcBundle if we try catch the BlockNotFound in PageBundle.

Because it is called from PageExtension (sonata page bundle) and it's called from twig, for this reason I got a twig exception 


And even if I create a new fielter tô check the privous exception, I'll skipp all blocks into the snapshot

VincentLanglet:
> Because it is called from PageExtension (sonata page bundle) and it's called from twig, for this reason I got a twig exception

I mean that you can still use this strategy https://github.com/sonata-project/SonataPageBundle/pull/1614 WITHOUT the pr on block bundle.
As a proof, tests are passing on your PR so you don't need this one.

jeromeengeln:
And use exists() function of blockRender before render to test if service exists ?

https://github.com/sonata-project/SonataPageBundle/blob/01f5a86138733dfbc28aafcf05d8dd52d88f4e8f/src/Twig/Extension/PageExtension.php#L189

```php
&& !$this->blockHelper->exists($block->getType()) 
```

eerison:
> > Because it is called from PageExtension (sonata page bundle) and it's called from twig, for this reason I got a twig exception
> 
> I mean that you can still use this strategy [sonata-project/SonataPageBundle#1614](https://github.com/sonata-project/SonataPageBundle/pull/1614) WITHOUT the pr on block bundle. As a proof, tests are passing on your PR so you don't need this one.

Do you mean use exception strategy in sonata page bundle?

VincentLanglet:
> Do you mean use exception strategy in sonata page bundle?

No. You really don't understand...

I mean that the build on your PR on PageBundle is Green, no ?
That means that no pr are needed on BlockBundle.

BlockBundle is already throwing a BlockNotFoundException when there is no Block, so you catch it and do nothing.
There is nothing related with the current PR which introduce a BlockServiceNotFound which is thrown in a different place than the one used by PageBundle.

VincentLanglet:
> And use exists() function of blockRender before render to test if service exists ?
> 
> https://github.com/sonata-project/SonataPageBundle/blob/01f5a86138733dfbc28aafcf05d8dd52d88f4e8f/src/Twig/Extension/PageExtension.php#L189
> 
> ```
> && !$this->blockHelper->exists($block->getType()) 
> ```

Yes, this is the perfect PR on PageBundle IMHO
And nothing is needed on BlockBundle.

eerison:
> BlockBundle is already throwing a BlockNotFoundException 

Nope, the current code is returning [RuntimeException](https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockServiceManager.php#L198)

> !$this->blockHelper->exists($block->getType()) 

IMO, we should just add a try catch, not check if the block type exist for each block in my page.


Well I can do this in SonataPageBundle, as I'm already doing, But we lost some benefits from Strategy Exception

for example, using strategyException I can create different filters to handle different scenarios like if debug is true or not, in dev environment is good to see the page crash and saying that my block doesn't exist and handle it just in prod.

But If I add a try catch in page bundle always the exception will be handled!

eerison:
> !$this->blockHelper->exists($block->getType()) 

it's not a good idea, It will hide completely the information that block doesn't exist! we need to log this error somehow!

VincentLanglet:
> > BlockBundle is already throwing a BlockNotFoundException
> 
> Nope, the current code is returning [RuntimeException](https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockServiceManager.php#L198)

Then I'd like the stacktrace, and not the twig one.

Because, inside the render method.

This call is supposed to only throw BlockNotFound
https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L390

There might be something wrong inside https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockContextManager.php#L128

And this is maybe https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockContextManager.php#L145
First, I dunno why we catch an ExceptionInterface.
But seems like the issue is that resolve https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockContextManager.php#L153 can throw the runtime exception from this call https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockContextManager.php#L287 because this method https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockServiceManager.php#L81 calls https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockServiceManager.php#L195.

So should we just change
```
    private function resolve(BlockInterface $block, array $settings): array
    {
        $optionsResolver = new OptionsResolver();

        $this->configureSettings($optionsResolver, $block);

        $service = $this->blockService->get($block);
        $service->configureSettings($optionsResolver);

        return $optionsResolver->resolve($settings);
    }
```
to
```
    private function resolve(BlockInterface $block, array $settings): array
    {
        $optionsResolver = new OptionsResolver();

        $this->configureSettings($optionsResolver, $block);

        try {
            $service = $this->blockService->get($block);
            $service->configureSettings($optionsResolver);
        } catch (\Exception) {
            // Todo
        }

        return $optionsResolver->resolve($settings);
    }
```
?

eerison:
Just one thing that came in my mind

why do we use `BlockHelper->render` instead of `BlockRenderer->render` ? looking the code BlockHelp uses BlockRenderer but add some cache code! is it used because of the cache?

VincentLanglet:
> Just one thing that came in my mind
> 
> why do we use `BlockHelper->render` instead of `BlockRenderer->render` ? looking the code BlockHelp uses BlockRenderer but add some cache code! is it used because of the cache?

Cache will be removed, but signature of the methods are different ; you have a block and not a blockContext.

src/Block/BlockServiceManager.php
```diff
@@ -191,11 +192,12 @@ public function validate(ErrorElement $errorElement, BlockInterface $block): voi
 
     /**
      * @throws \RuntimeException
+     * @throws BlockServiceNotFoundException
      */
     private function load(string $type): BlockServiceInterface
     {
         if (!$this->has($type)) {
-            throw new \RuntimeException(sprintf('The block service `%s` does not exist', $type));
+            throw new BlockServiceNotFoundException(sprintf('The block service `%s` does not exist', $type));
```
VincentLanglet, line 199: Can you also throw a new BlockServiceNotFoundException line 206 ?

src/Block/BlockServiceManager.php
```diff
@@ -191,11 +192,12 @@ public function validate(ErrorElement $errorElement, BlockInterface $block): voi
 
     /**
      * @throws \RuntimeException
```
VincentLanglet:
```suggestion
```

eerison:
> > Just one thing that came in my mind
> > why do we use `BlockHelper->render` instead of `BlockRenderer->render` ? looking the code BlockHelp uses BlockRenderer but add some cache code! is it used because of the cache?
> 
> Cache will be removed, but signature of the methods are different ; you have a block and not a blockContext.

ahhhhh I got it! for this reason it's getting the context here https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L390

well But IMO make sense to add a try catch here: https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L390 I don't see a problem to add a try/catch there 👀

VincentLanglet:
> well But IMO make sense to add a try catch here: https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L390 I don't see a problem to add a try/catch there 👀

We can catch it directly here: https://github.com/sonata-project/SonataBlockBundle/pull/1117
Does it solve your issue ?

eerison:
```
private function resolve(BlockInterface $block, array $settings): array
    {
        $optionsResolver = new OptionsResolver();

        $this->configureSettings($optionsResolver, $block);

        try {
            $service = $this->blockService->get($block);
            $service->configureSettings($optionsResolver);
        } catch (\Exception) {
            // Todo
        }

        return $optionsResolver->resolve($settings);
    }
```

well I can add the try catch here, But the point is: Can I add the catch like this: https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockRenderer.php#L73-L86 👀

eerison:
> > well But IMO make sense to add a try catch here: https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L390 I don't see a problem to add a try/catch there 👀
> 
> We can catch it directly here: #1117 Does it solve your issue ?

it solves, But the info that BlockService doesn't exist will be hidden,

as I said above, can we add this

https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockRenderer.php#L73-L86

eerison:
> > > well But IMO make sense to add a try catch here: https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L390 I don't see a problem to add a try/catch there 👀
> > 
> > 
> > We can catch it directly here: #1117 Does it solve your issue ?
> 
> it solves, But the info that BlockService doesn't exist will be hidden,
> 
> as I say above, can we add this
> 
> https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockRenderer.php#L73-L86

and other thing, if `Block` and `BlockService` are different, then `BlockServiceNotFoundException` should not implements `BlockNotFoundInterface`

VincentLanglet:
> > > well But IMO make sense to add a try catch here: https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L390 I don't see a problem to add a try/catch there 👀
> > 
> > 
> > We can catch it directly here: #1117 Does it solve your issue ?
> 
> it solves, But the info that BlockService doesn't exist will be hidden,
> 
> as I said above, can we add this
> 
> https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockRenderer.php#L73-L86

I can add log like this https://github.com/sonata-project/SonataBlockBundle/pull/1117/commits/17f93fcba93d1a913a2dd4de3ac45ac2edca5e01

But the exceptionManagerStrategy should not be used since it's supposed to return a Response and we won't use it.

eerison:
> > > > well But IMO make sense to add a try catch here: https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L390 I don't see a problem to add a try/catch there 👀
> > > 
> > > 
> > > We can catch it directly here: #1117 Does it solve your issue ?
> > 
> > 
> > it solves, But the info that BlockService doesn't exist will be hidden,
> > as I said above, can we add this
> > https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockRenderer.php#L73-L86
> 
> I can add log like this [17f93fc](https://github.com/sonata-project/SonataBlockBundle/commit/17f93fcba93d1a913a2dd4de3ac45ac2edca5e01)
> 
> But the exceptionManagerStrategy should not be used since it's supposed to return a Response and we won't use it.

Yeah it is a bit better then do not log anything 😄 

But I guess it should not require exception, it should only handle exception and nothing more!

eerison:
> > > > > well But IMO make sense to add a try catch here: https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Templating/Helper/BlockHelper.php#L390 I don't see a problem to add a try/catch there 👀
> > > > 
> > > > 
> > > > We can catch it directly here: #1117 Does it solve your issue ?
> > > 
> > > 
> > > it solves, But the info that BlockService doesn't exist will be hidden,
> > > as I said above, can we add this
> > > https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockRenderer.php#L73-L86
> > 
> > 
> > I can add log like this [17f93fc](https://github.com/sonata-project/SonataBlockBundle/commit/17f93fcba93d1a913a2dd4de3ac45ac2edca5e01)
> > But the exceptionManagerStrategy should not be used since it's supposed to return a Response and we won't use it.
> 
> Yeah it is a bit better then do not log anything 😄
> 
> But I guess it should not require exception, it should only handle exception and nothing more!

But well it's requiring in the interface and it'll be too much work for now :/
https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Exception/Strategy/StrategyManagerInterface.php#L33

VincentLanglet (changes requested):

VincentLanglet:
As explained here: https://github.com/sonata-project/SonataBlockBundle/pull/1117#issuecomment-1256019793, this PR works fine with the config 
```
sonata_block:
    default_contexts: [sonata_page_bundle]
    exception:
        default:
            filter: ignore_block_exception
```

Also, by default it's debug only ; so it won't throw any exception in production. So I think this is not a real issue.

eerison:
> Also, by default it's debug only ; so it won't throw any exception in production. So I think this is not a real issue.

But the page doesn't render the others blocks

eerison:
I'm using the same config that you added above and my page is still broken

VincentLanglet (dismissed):

VincentLanglet:
> > Also, by default it's debug only ; so it won't throw any exception in production. So I think this is not a real issue.
> 
> But the page doesn't render the others blocks

Yes but this is not an issue.
You decided it should display other blocks but why ? The page might not have any sens anymore if you hide one block.
Having this default behavior is not choking ; then people can change the strategy, like I did

VincentLanglet:
@JeromeEngelnAdeliom Can you update the tests too ? =)

eerison:
> The page might not have any sens anymore if you hide one block.

do you mean if one block fail, my whole page no make sense anymore? I don't get it!

VincentLanglet:
> > The page might not have any sens anymore if you hide one block.
> 
> do you mean if one block fail, my whole page no make sense anymore? I don't get it!

Yes it can.

Let's say you have a block Text "Please subscribe by giving your email on the next line".
Then you display a Block Email but this one doesn't exist (so is not displayed).
The first block alone doesn't make any sens.

I can find more and more example. So the current beahvior is understandable ; and if you want a different one you'll be able to use another manager strategy. Imho it require just some doc on SonataPageBundle.

src/Exception/BlockServiceNotFoundException.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\BlockBundle\Exception;
+
+final class BlockServiceNotFoundException extends \Exception implements BlockExceptionInterface
```
VincentLanglet:
```suggestion
final class BlockServiceNotFoundException extends \RuntimeException implements BlockExceptionInterface
```

VincentLanglet (commented):
The fact that the tests failed mean we introduced a regression. To avoid this we need that BlockServiceNotFoundException extends RuntimeException

VincentLanglet (approved):

eerison:
> > > The page might not have any sens anymore if you hide one block.
> > 
> > 
> > do you mean if one block fail, my whole page no make sense anymore? I don't get it!
> 
> Yes it can.
> 
> Let's say you have a block Text "Please subscribe by giving your email on the next line". Then you display a Block Email but this one doesn't exist (so is not displayed). The first block alone doesn't make any sens.
> 
> I can find more and more example. So the current beahvior is understandable ; and if you want a different one you'll be able to use another manager strategy. Imho it require just some doc on SonataPageBundle.

Well it really depends of each case

in my case some pages uses blocks completely independent each other, and if one block fail, make totally sense for me still keep my page showing with others blocks. But if it's not an impediment for sonata page 4 then It's fine

VincentLanglet:
> in my case some pages uses blocks completely independent each other, and if one block fail, make totally sense for me still keep my page showing with others blocks.

And because it depends of each case, each developper should be able to decide what to do.
AND that the job of the configuration of the exception manager strategy. I dunno how many times I have to say it.

jordisala1991 (approved):
The PR looks ok to me, but I can't follow all the conversation. I let you decide if this is ready to merge @VincentLanglet

VincentLanglet:
> The PR looks ok to me, but I can't follow all the conversation. I let you decide if this is ready to merge @VincentLanglet

To me, it is. And this works fine for me. Dunno why @eerison  doesn't made it work for him.

PR #1121, opened 2022-10-09
Fix init of block settings

core23:
## Subject

This is a bug fix for the usage without the page bundle.

```twig
    {{ sonata_block_render({ 'type': 'sonata.media.block.gallery' }, {
        // ...
        "galleryId": 2
    }) }}
```

When passing a block definition `$meta` array to this method, the `$settings` will be ignored for this block itself, they are only passed to the block context. This will result in a null value for different blocks (e.g. https://github.com/sonata-project/SonataMediaBundle/blob/9946d8ac3e0f21f1198050e5f7501766b058e53a/src/Block/GalleryBlockService.php#L169).

## Changelog

```markdown
### Fixed
- Fix init of block settings
```

### DIFF
--- a/src/Block/BlockContextManager.php
+++ b/src/Block/BlockContextManager.php
@@ -134,6 +134,8 @@ final class BlockContextManager implements BlockContextManagerInterface
                 // merge user settings
                 $settings = array_merge($meta['settings'], $settings);
             }
+
+            $block->setSettings($settings);
         } else {
             $block = $meta;
         }
--- a/tests/Block/BlockContextManagerTest.php
+++ b/tests/Block/BlockContextManagerTest.php
@@ -21,6 +21,7 @@ use Sonata\BlockBundle\Block\BlockContextManager;
 use Sonata\BlockBundle\Block\BlockLoaderInterface;
 use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
 use Sonata\BlockBundle\Block\Service\AbstractBlockService;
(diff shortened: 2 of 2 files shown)


src/Block/BlockContextManager.php
```diff
@@ -134,6 +134,8 @@ public function get($meta, array $settings = []): BlockContextInterface
                 // merge user settings
                 $settings = array_merge($meta['settings'], $settings);
             }
+
+            $block->setSettings($settings);
```
VincentLanglet, line 138: Are we supposed to also set the Settings in the `else` part, this would be consistent no ?

core23: Not sure about this. When you pass a `BlockInterface` in the else part your block could already have valid settings.

VincentLanglet: We doing later
```
try {
            $settings = $this->resolve($block, array_merge($block->getSettings(), $settings));
        } catch (ExceptionInterface $e) {
```
So this means that the $settings might not be compatible with the Block provided.

That's why I wonder if we shouldn't set the settings after the try/catch call

VincentLanglet:
@jordisala1991 You worked a little with blocks when upgrading SonataPageBundle, you might know this better than me

VincentLanglet:
Friednly ping @jordisala1991

jordisala1991 (approved):

VincentLanglet (approved):

PR #1155, opened 2023-03-20
Remove doctrine common

jordisala1991:
## Subject

Closes #{put_issue_number_here}.

## Changelog

```markdown
### Added
- Added some `Class::newMethod()` to do great stuff.

### Changed

### Deprecated

### Removed

### Fixed

### Security
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -67,7 +67,6 @@
         "vimeo/psalm": "^4.1 || ^5.8.0"
     },
     "conflict": {
-        "doctrine/annotations": "<1.7",
         "knplabs/knp-menu-bundle": "<3.0"
     },
     "suggest": {
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,6 @@
     "require": {
         "php": "^8.0",
         "doctrine/collections": "^1.8 || ^2.0",
-        "doctrine/common": "^2.7.1 || ^3.0",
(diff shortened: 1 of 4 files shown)


jordisala1991:
Not sure if it is safe to do this change or we have to do it different.

jordisala1991:
Wdyt @VincentLanglet ?

VincentLanglet (commented):
I'm not sure about it, ClassUtils was used for Proxy classes.
If the blocks are doctrine entities, it may won't work.

In SonataPage we have blocks as entities 
https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Entity/BaseBlock.php
so this might be an issue here...

I dunno what's the doctrine recommendation...

jordisala1991:
I think with the new proxy system on the orm it might not be an issue anymore, but it is only an opt in on 2.14 sf 6.2 and persistence 3.1, it might not be safe indeed.

Another option is to port the code to avoid the dep, wdyt?

VincentLanglet:
To port this in BlockBundle we will still need to add `doctrine-persistence` as a dependency.

If we do this, we could follow the following suggestion: https://github.com/doctrine/persistence/pull/326#issuecomment-1374466942

jordisala1991:
I think we should wait more for this change.

PR #1158, opened 2023-03-27
Remove serializer configuration

jordisala1991:
## Subject

I am targeting this branch, because I guess this can be considered BC break.

## Changelog

```markdown
### Removed
- Remove block serializer configuration
```


jordisala1991:
To me it looks that since the v4 of page-bundle this should not make sense (we removed the api on all bundles, and page bundle was the only one exposing blocks)

VincentLanglet (approved):

PR #1159, opened 2023-03-27
Add return types to FormMapper methods

jordisala1991:
## Subject

I am targeting this branch, because this is in preparation for 5.0.

## Changelog

```markdown
### Added
- Add return types to `FormMapper`.
```


VincentLanglet (approved):
It doesn't break the compatibility with https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Form/FormMapper.php

PR #1162, opened 2023-04-19
Fix phpstan

jordisala1991:


jordisala1991:
TBH I don't know if this is the correct fix or how should we know what is the generic type of those.... cc @VincentLanglet this will help with: https://github.com/sonata-project/SonataAdminBundle/pull/8044

src/Form/Mapper/FormMapper.php
```diff
@@ -39,8 +39,8 @@ public function create(string $name, ?string $type = null, array $options = []):
     public function reorder(array $keys);
 
     /**
-     * @param class-string<FormTypeInterface>|null $type
-     * @param array<string, mixed>                 $options
+     * @param class-string<FormTypeInterface<mixed>>|null $type
```
VincentLanglet: This will create a covariant issue:
https://phpstan.org/r/a245eba4-e65b-4d08-8a99-ab8897f4eca3

A right solution would be to make the FormMapper generic, but that can be anoying for us having to precise the template everywhere...

Another solution could be to use the `skipCheckGenericClass` feature https://github.com/phpstan/phpstan-symfony/blob/1.2.x/extension.neon#L15

jordisala1991: This skip generic classes is a problem for tomorrow. Next major will empty that array and trigger those errors again.

The covariant issue is already there if I try to define the type on the admin-bundle.

jordisala1991: not sure about what you mean making this class generic. The form mapper add method accepts any formType, I dont think we can make it generic.

jordisala1991: I made both methods generic (not the class): https://phpstan.org/r/bb377072-cebf-408b-95e1-e81a65036e7a

PR #1229, opened 2025-05-02
fixes BlockServiceTestCase

mvhirsch:
## Subject

Looks like the docs are outdated, as there is no such `AbstractBlockServiceTestCase`.

I weren't able to `execute()` my block either, because `Twig` get's mocked. Not sure what to test though. That's why I suggest removing these lines.

I am targeting this branch, because its a patch.

~~Closes #{put_issue_number_here}.~~

### DIFF
--- a/docs/reference/testing.rst
+++ b/docs/reference/testing.rst
@@ -32,9 +32,9 @@ Given the following block service::
 
 You can write unit tests for block services with the following code::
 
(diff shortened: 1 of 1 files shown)


VincentLanglet (approved):

VincentLanglet:
Thanks

PR #1247, opened 2025-11-25
Fix symfony 7.3 deprecation in RssBlockService

eerison:
## Fix symfony 7.3 deprecation in RssBlockService

Since symfony/validator 7.3: Passing an array of options to configure the "Symfony\Component\Validator\Constraints\Length" constraint is deprecated, use named arguments instead.

## Changelog

```markdown
### Changed
- Changed passing `array` to `named` argument for `Length` validation constraint in `src/Block/Service/RssBlockService.php`

### Added
- Added `default_protocol` equals `true` for UrlType` at `src/Block/Service/RssBlockService.php`
```

Related with: https://github.com/sonata-project/SonataPageBundle/issues/1837

### DIFF
--- a/src/Block/Service/RssBlockService.php
+++ b/src/Block/Service/RssBlockService.php
@@ -94,7 +94,7 @@ final class RssBlockService extends AbstractBlockService implements EditableBloc
             ->with('settings[title]')
                 ->addConstraint(new NotNull())
                 ->addConstraint(new NotBlank())
-                ->addConstraint(new Length(['max' => 50]))
+                ->addConstraint(new Length(max: 50))
             ->end();
     }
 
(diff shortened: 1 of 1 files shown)


eerison:
there are others depredations, I will try to add in this PR

src/Block/Service/RssBlockService.php
```diff
@@ -56,6 +56,7 @@ public function configureEditForm(FormMapper $form, BlockInterface $block): void
             'keys' => [
                 ['url', UrlType::class, [
                     'required' => false,
+                    'default_protocol' => 'https',
```
eerison, line 59: on symfony 7.3 the default value is `http`: https://symfony.com/doc/current/reference/forms/types/url.html#default-protocol

But I guess we could keep `https` as default, WDYT?

eerison:
> there are others depredations, I will try to add in this PR

I just found those 2 deprecation.

src/Block/Service/RssBlockService.php
dmaicher: works for me :+1:

dmaicher (approved):

VincentLanglet (approved):

